PR #8, opened 2020-03-17
Switch to new client API

alexshtin:
1. `NewClient` doesn't accept `workflowservice`.
2. `hostPort` is optional.
2. `domain` is optional.


cmd/samples/cron/main.go
```diff
@@ -1,66 +1,125 @@
...
 	"flag"
 	"time"
 
 	"github.com/pborman/uuid"
+	"go.temporal.io/temporal-proto/serviceerror"
+	"go.temporal.io/temporal-proto/workflowservice"
 	"go.temporal.io/temporal/client"
 	"go.temporal.io/temporal/worker"
-
-	"github.com/temporalio/temporal-go-samples/cmd/samples/common"
```
alexshtin: This is self contained sample w/o any dependencies. @mfateev, if you are ok with this approach, I will refactor all the rest.

cmd/samples/cron/main.go
```diff
@@ -1,66 +1,125 @@
...
+	"go.uber.org/zap"
 )
 
 const (
-	// ApplicationName is the task list for this sample
-	ApplicationName = "cronGroup"
+	// DomainName is the name of domain where workflow will be created.
+	DomainName = "samples-domain"
```
alexshtin: No config. `DomainName` is hardcoded, and SDK default `hostPort` is used.

cmd/samples/cron/main.go
```diff
@@ -1,66 +1,125 @@
...
-		// Use select{} to block indefinitely for samples, you can quit by CMD+C.
+		// Use select{} to block indefinitely for samples, you can quit by Ctrl+C.
 		select {}
 	case "trigger":
-		startWorkflow(&h, cron)
+		startWorkflow(cron)
+	}
+}
+
+func createLogger() {
```
alexshtin: Will copy `createLogger` and `createDomain` to every sample.

cmd/samples/cron/main.go
```diff
@@ -1,66 +1,125 @@
...
+}
+
+func createDomain() {
+	domainClient, err := client.NewDomainClient(client.Options{})
+
+	if err != nil {
+		logger.Fatal("Unable to create domain client", zap.Error(err))
+	}
+
+	err = domainClient.Register(context.Background(), &workflowservice.RegisterDomainRequest{
```
alexshtin: For some reason it was just `DescribeDomain`. I don't see any issues creating domain right from here.

cmd/samples/cron/main.go
mfateev: Let's make domain optional in the sdks with defaulting it to "default". Then this is not necessary anymore.

cmd/samples/cron/main.go
mfateev: Why do we have createDomain at all?

mfateev: I would not have these methods. Just put logger creaton into main

cmd/samples/common/factory.go
```diff
@@ -30,9 +26,9 @@ func NewBuilder(logger *zap.Logger) *WorkflowClientBuilder {
 	}
 }
 
-// SetHostPort sets the hostport for the builder
-func (b *WorkflowClientBuilder) SetHostPort(hostport string) *WorkflowClientBuilder {
-	b.hostPort = hostport
+// SetHostPort sets the host:port for the builder
+func (b *WorkflowClientBuilder) SetHostPort(hostPort string) *WorkflowClientBuilder {
```
samarabbas: Do we really need this builder abstraction anymore?

cmd/samples/cron/main.go
```diff
@@ -1,66 +1,95 @@
...
+
 // This needs to be done as part of a bootstrap step when the process starts.
 // The workers are supposed to be long running.
-func startWorkers(h *common.SampleHelper) {
-	// Configure worker options.
+func startWorker() {
 	workerOptions := worker.Options{
-		MetricsScope: h.Scope,
-		Logger:       h.Logger,
+		Logger: logger,
```
samarabbas: How do we configure hostport?  If someone wants to run samples against a real deployment of Temporal.

cmd/samples/cron/main.go
```diff
@@ -1,66 +1,95 @@
...
-	flag.StringVar(&cron, "cron", "* * * * *", "Crontab schedule. Default \"* * * * *\"")
-	flag.Parse()
 
-	var h common.SampleHelper
-	h.SetupServiceConfig()
-
-	switch mode {
-	case "worker":
-		startWorkers(&h)
+	client, err := client.NewClient(DomainName, client.Options{})
```
samarabbas: We should have a an easy config based mechanism to target samples against any deployment of Temporal cluster.

cmd/samples/dsl/main.go
```diff
@@ -22,7 +22,13 @@ func startWorkers(h *common.SampleHelper) {
 		MetricsScope: h.Scope,
 		Logger:       h.Logger,
 	}
-	h.StartWorkers(h.Config.DomainName, ApplicationName, workerOptions)
+	worker := h.StartWorker(h.Config.DomainName, ApplicationName, workerOptions)
+	worker.RegisterWorkflow(SimpleDSLWorkflow)
```
samarabbas: Workflow and activities should be registered before calling start.

cmd/samples/common/factory.go
alexshtin: No. Entire `common` will go away when I complete all other samples. I refactored only `cron` so far.

cmd/samples/cron/main.go
alexshtin: I don't want to use config because it adds complexity to the code and restriction to how we run samples. I can pass default value `localhost:7233` using options and it will allow users to change it easily.

cmd/samples/dsl/main.go
alexshtin: Yeah, I noticed this also. So none of these samples (besides `cron`) actually work now.

cmd/samples/cron/main.go
alexshtin: I've just created `client.DefaultHostPort` const in SDK. I will pass it explicitly here.

cmd/samples/cron/main.go
```diff
@@ -1,66 +1,108 @@
...
-	ApplicationName = "cronGroup"
+	// TaskListName is the task list for this sample.
+	TaskListName = "cron-task-list"
+)
+
+var (
+	logger *zap.Logger
 )
 
+func main() {
```
alexshtin: Moved `main` on top.

cmd/samples/cron/main.go
```diff
@@ -1,66 +1,108 @@
...
+	switch mode {
+	case "worker":
+		w := startWorker()
+		// The workers are supposed to be long running process that should not exit.
+		// Use channel to wait for Ctrl+C.
+		c := make(chan os.Signal, 1)
+		signal.Notify(c, os.Interrupt)
+		<-c
+		// Stop worker, close connection, clean up resources.
+		w.Stop()
```
alexshtin: Added two extra lines (and some complexity) but I believe it is good to show that workers must/can be stopped.

cmd/samples/cron/main.go
```diff
@@ -1,66 +1,108 @@
...
+	case "trigger":
+		startWorkflow(cron)
+	default:
+		flag.PrintDefaults()
```
alexshtin: This will print user friendly help screen.

cmd/samples/cron/main.go
```diff
@@ -25,29 +22,70 @@ var (
...
-		logger.Error("Unable to start worker", zap.Error(err))
+		logger.Fatal("Unable to start worker", zap.Error(err))
 	}
+
+	return w
 }
 
 // Start instance of the workflow.
 func startWorkflow(cron string) {
+	c, err := client.NewClient(client.Options{
```
mfateev: Add comment that the client is a heavyweight object and should be created once per process.

cmd/samples/cron/main.go
```diff
@@ -25,29 +22,70 @@ var (
...
+		panic(err)
+	}
+
+	var mode string
+	var cron string
+	flag.StringVar(&mode, "m", "", "Mode is worker or trigger.")
+	flag.StringVar(&cron, "cron", "* * * * *", "Crontab schedule.")
+	flag.Parse()
+
+	switch mode {
```
mfateev: What about creating two mains? This would eliminate most of the plumbing code.

cron/starter/main.go
```diff
@@ -0,0 +1,56 @@
...
+	var err error
+	logger, err = zap.NewDevelopment()
+	if err != nil {
+		panic(err)
+	}
+
+	startWorkflow()
+}
+
+func startWorkflow() {
```
mfateev: Remove the function. Just do it in main

cron/starter/main.go
```diff
@@ -0,0 +1,56 @@
+package main
```
alexshtin: Two mains allow just to run `go run main.go` in corresponding dirrecotry.

cron/worker/main.go
```diff
@@ -0,0 +1,64 @@
...
+	// The workers are supposed to be long running process that should not exit.
+	c, w := startWorker()
+	waitCtrlC()
+	// Stop worker, close connection, clean up resources.
+	w.Stop()
+	_ = c.CloseConnection()
+}
+
+// This needs to be done as part of a bootstrap step when the process starts.
+func startWorker() (client.Client, worker.Worker) {
```
mfateev: Merge function into main

cron/worker/main.go
```diff
@@ -0,0 +1,64 @@
...
+	})
+	if err != nil {
+		logger.Fatal("Unable to create client", zap.Error(err))
+	}
+
+	workerOptions := worker.Options{
+		Logger: logger,
+	}
+
+	w := worker.New(c, "cron-task-list", workerOptions)
```
mfateev: Either initialize both option structures inline or both in a separate variable

cron/worker/main.go
```diff
@@ -0,0 +1,65 @@
...
+	if err != nil {
+		panic(err)
+	}
+
+	// The workers are supposed to be long running process that should not exit.
+	c, w := startWorker()
+	waitCtrlC()
+	// Stop worker, close connection, clean up resources.
+	w.Stop()
+	_ = c.CloseConnection()
```
alexshtin: Example of  how to cleanup properly.

cron/worker/main.go
alexshtin: Just realized this myself!

cron/starter/main.go
alexshtin: Was in doubt about this.

PR #19, opened 2020-04-28
Fixed fileprocessing

mfateev:
Fix for https://github.com/temporalio/temporal-go-samples/issues/17

Refactored the fileprocessing sample as it was completely broken after the session was added to it.

### DIFF
--- a/fileprocessing/worker/main.go
+++ b/fileprocessing/worker/main.go
@@ -18,7 +18,7 @@ func main() {
 		panic(err)
 	}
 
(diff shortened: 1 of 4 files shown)


fileprocessing/worker/main.go
```diff
@@ -18,7 +18,7 @@ func main() {
 		panic(err)
 	}
 
-	// The client and worker are heavyweight objects that should be created once per process.
+	// The client and w are heavyweight objects that should be created once per process.
```
samarabbas, line 21: misspelled.

PR #22, opened 2020-05-11
Add signal to childworkflow example

samarabbas:
`childworkflow` example uses signal API to signal child workflow.


alexshtin:
I am closing this as outdated.

PR #26, opened 2020-05-27
Update dependencies

alexshtin:
Update go SDK and proto. This leads to follow changes:
1. `CloseConnection` -> `Close`.
2. `worker.Start()` -> `worker.Run()`.
3. `LICENSE` file matches server one.


LICENSE
```diff
@@ -1,13 +1,23 @@
-Copyright (c) 2017 Uber Technologies, Inc.
+The MIT License
```
mfateev, line 1: Let's keep the same Apache license. The relicensing should be done more carefully and only if needed.

LICENSE
```diff
@@ -1,13 +1,23 @@
-Copyright (c) 2017 Uber Technologies, Inc.
```
rylandg, line 1: Is this legal?

PR #71, opened 2020-12-05
Capture err value when activity returns

cuminandpaprika:
Currently the `recordActivityEnd` function evaluates the value of the err parameter when the defer statement is called, rather than when the activity returns. This means that any errors produced by the activity won't be correctly recorded.

For a simple example demonstrating the current behavior, please refer to https://play.golang.org/p/7gVYcXJla8E


alexshtin (approved):
Thanks for fixing this!

PR #86, opened 2021-03-16
Add an example of encrypting payloads with a DataConverter

robholland:


encrypted-payloads/crypt-converter.go
```diff
@@ -0,0 +1,150 @@
...
+	return result
+}
+
+// ToString converts payload object into human readable string.
+func (dc *CryptDataConverter) ToString(payload *commonpb.Payload) string {
+	decryptData, err := decrypt(payload.GetData(), dc.getKey())
+	// No way to return an error here...
+	if err == nil {
+		payload.Data = decryptData
+	}
```
alexshtin, line 147:
```suggestion
if err != nil {
	return err.Error()
}
payload.Data = decryptData
```

encrypted-payloads/crypt-converter.go
```diff
@@ -0,0 +1,150 @@
...
+}
+
+// NewCryptDataConverter created new instance of CryptDataConverter wrapping a DataConverter
+func NewCryptDataConverter(dataConverter converter.DataConverter) *CryptDataConverter {
+	return &CryptDataConverter{
+		dataConverter: dataConverter,
+	}
+}
+
+func encrypt(plaintext []byte, key []byte) ([]byte, error) {
```
alexshtin, line 33:
```suggestion
func encrypt(rawData []byte, key []byte) ([]byte, error) {
```

encrypted-payloads/crypt-converter.go
```diff
@@ -0,0 +1,150 @@
...
+
+	nonce := make([]byte, gcm.NonceSize())
+	if _, err = io.ReadFull(rand.Reader, nonce); err != nil {
+		return nil, err
+	}
+
+	return gcm.Seal(nonce, nonce, plaintext, nil), nil
+}
+
+func decrypt(ciphertext []byte, key []byte) ([]byte, error) {
```
alexshtin, line 52:
```suggestion
func decrypt(cipherData []byte, key []byte) ([]byte, error) {
```

encrypted-payloads/crypt-converter.go
alexshtin: or `rawBytes`, or `plainData`. It is not "text".

encrypted-payloads/README.md
```diff
@@ -0,0 +1,10 @@
+### Steps to run this sample:
+1) You need a Temporal service running. See details in README.md
+2) Run the following command to start the worker
+```
+go run worker/main.go
+```
+3) Run the following command to start the example
+```
+go run starter/main.go
+```
```
alexshtin, line 10: 4) Run
```
tctl workflow show --wid encrypted-payloads_workflowID
```
and observer encrypted agrs in the history.

alexshtin: Please, just run it yourself first to make sure it works.

encrypted-payloads/crypt-converter.go
```diff
@@ -0,0 +1,150 @@
+package cryptconverter
```
alexshtin, line 1: I don't like package name is not the same as dir name. Move this to `cryptconverter` sub-dir and remove package alias in the worker and starter.

alexshtin (approved):

encrypted-payloads/crypt-converter.go
Sushisource: Just FYI I would find this weird. I have never seen anyone call it `cypherData`, even if maybe that makes more sense. It's always `cypherText` IME -- or `encrypted` or `encryptedData`, which is probably what I'd go with here.

encrypted-payloads/README.md
robholland: This will show:
```
...
  11  WorkflowExecutionCompleted  {Result:[encoding
                                  binary/crypt: payload
                                  encoding is not supported],
                                  WorkflowTaskCompletedEventId:10}
```

So it won't actually show the raw bytes here, but I think that's ok?

robholland: An alternative would be to mark this as a byteslice payload and use `metadata["wrapped-encoding"]` to detect if we should decrypt. That way `tctl` would show the bytes. What do you think?

robholland: Switched to marking the encrypted payload as `binary/plain` so that the tools display the bytes.

cryptconverter/data-converter.go
```diff
@@ -0,0 +1,183 @@
...
+	"io"
+
+	commonpb "go.temporal.io/api/common/v1"
+
+	"go.temporal.io/sdk/converter"
+)
+
+const (
+	// MetadataWrappedEncoding is "wrapped-encoding"
+	MetadataWrappedEncoding = "wrapped-encoding"
```
robholland, line 17: Not sure about this name, welcome opinions.

Sushisource (approved):
LGTM!

sw-yx:
ty!

PR #95, opened 2021-05-07
Fix openNClosed fixture not completing activities

feedmeapples:
## What was changed:

Matched task queues of worker and starter

## Why?

addresses worker that was listening on another task queue  

## Checklist

1. Closes issue: 

2. How was this tested:

ran the fixture

3. Any docs updates needed?

no


sw-yx (approved):
oops

PR #112, opened 2021-06-22
Add ZapAdapter sample

alexshtin:
Add `ZapAdapter` and its usage with `zap.Logger` .


zapadapter/worklfow.go
```diff
@@ -0,0 +1,51 @@
...
+import (
+	"context"
+	"errors"
+	"time"
+
+	"go.temporal.io/sdk/activity"
+	"go.temporal.io/sdk/workflow"
+)
+
+// Workflow is a Hello World workflow definition.
```
Sushisource, line 12: Comment needs an update. Just a brief explanation that it just does a few different kinds of logging is fine.

zapadapter/worker/main.go
```diff
@@ -0,0 +1,69 @@
...
+	"go.temporal.io/sdk/worker"
+	"go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
+)
+
+func main() {
+	// The client and worker are heavyweight objects that should be created once per process.
+	c, err := client.NewClient(client.Options{
+		Logger: zapadapter.NewZapAdapter(NewZapLogger()),
+	})
```
Sushisource, line 17: Probably worth calling attention to this with a more explicit comment since it's the key part

Sushisource (approved):

PR #128, opened 2021-08-11
Signal only running workflows in workflows fixture

feedmeapples:
## What was changed

Signals only running workflows in a fixture, not all 

## Why?

Fixes an issue that the workflows fixture would fail before starting all of the workflows

## Checklist

1. Closes 

2. How was this tested:

ran the fixture on with `NumberOfSets` = 5
verified workflows are started and have expected statuses through web ui
verified only running workflwos have signal event

3. Any docs updates needed?

no


sw-yx (approved):
now it works!

PR #133, opened 2021-08-29
Updatabletimer

mfateev:
## What was changed
Added `updatabletimer` example.

## Why?
Java SDK has this example for a while. It implements a sleep function that can be rescheduled through an external signal at any time. It demonstrates use of Selector to wait on both a Channel and a Future simultaneously. It also demonstrates timer cancellation through a `workflow.Context` passed to the `workflow.NewTimer`.

2. How was this tested:
The sample includes a unit test. Its three executables were also tested manually.

3. Any docs updates needed?
The sample contains its own README. The root project README is also updated to include the new sample.


tsurdilo (approved):

tsurdilo:
lgtm. just needs rebase

PR #134, opened 2021-09-06
Add a workflow cancellation check

agbaraka:
## What was changed
This change adds a workflow cancellation check in the defer function 

## Why?
The "cleanup" activity in the defer function runs every time when the workflow completes, whether is has been cancelled or NOT.

The check added ensures that the activity in the defer function only runs when the workflow is cancelled.

## Checklist
1. How was this tested:
- Cancelling a long running workflow to see if the cleanup activity will be ran
- Letting a workflow to complete and see if the cleanup activity will be triggered

2. Any docs updates needed? 
https://docs.temporal.io/docs/go/workflows#how-to-clean-up-after-a-workflow-is-cancelled


alexshtin (approved):

tsurdilo:
@agbaraka this pr needs to be rebased. Thanks

PR #140, opened 2021-10-01
Fix encrypted-payloads data converter behaviour for local activity results

robholland:
Fixes a bug which caused local activity results to not be encrypted in the history.

Also avoids trying to decrypt a payload that hasn't been encrypted.

### DIFF
--- a/encrypted-payloads/crypt-converter.go
+++ b/encrypted-payloads/crypt-converter.go
@@ -131,12 +131,7 @@ func (dc *CryptDataConverter) ToPayloads(values ...interface{}) (*commonpb.Paylo
 			return nil, fmt.Errorf("values[%d]: %w", i, err)
 		}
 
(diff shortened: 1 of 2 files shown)


encrypted-payloads/crypt-converter.go
```diff
@@ -118,13 +174,6 @@ func (dc *CryptDataConverter) FromPayloads(payloads *commonpb.Payloads, valuePtr
 
 // FromPayload converts single value from payload.
 func (dc *CryptDataConverter) FromPayload(payload *commonpb.Payload, valuePtr interface{}) error {
-	decryptData, err := decrypt(payload.GetData(), dc.getKey())
-	if err != nil {
-		return fmt.Errorf("%w: %v", converter.ErrUnableToDecode, err)
-	}
-
-	payload.Data = decryptData
```
yiminc, line 126: this is the actual bug, overwrite the input memory block.

yiminc (approved):

PR #157, opened 2021-12-08
Update SDK version to v1.12.0

cretz:
## What was changed

Update SDK version to v1.12.0 which changed how Tally and OpenTracing are used a bit.

## Why?

Part of normal upgrade cycle


ctxpropagation/starter/main.go
```diff
@@ -4,23 +4,29 @@ import (
...
-	"github.com/temporalio/samples-go/ctxpropagation"
 )
 
 func main() {
 	// Set tracer which will be returned by opentracing.GlobalTracer().
 	closer := ctxpropagation.SetJaegerGlobalTracer()
 	defer func() { _ = closer.Close() }()
 
+	// Create interceptor
+	tracingInterceptor, err := opentracing.NewInterceptor(opentracing.TracerOptions{})
```
mmcshane, line 21: One to keep in the back of your mind - a set of `MustNewInterceptor` functions might be useful to clients.

metrics/metrics.go
```diff
@@ -15,21 +15,24 @@ const (
 	activitySuccessCount = "activity_succeeded"
 )
 
-func recordActivityStart(scope tally.Scope, activityType string, scheduledTimeNanos int64) (tally.Scope, tally.Stopwatch) {
-	scope = scope.Tagged(map[string]string{"operation": activityType})
+func recordActivityStart(
+	handler client.MetricsHandler,
+	activityType string,
+	scheduledTimeNanos int64,
+) client.MetricsHandler {
```
mmcshane, line 22: _Slightly_ confusing to return `metrics.MetricsHandler` here (though reading the code - I get it). Any opinion on returning a `func(error)` that closes over the handler and the startTime?

mmcshane (approved):
Works for me. One suggested but minor improvement and one musing-for-the-future.

metrics/metrics.go
alexshtin: This is samples repo. All code here should be straight forward as much as possible. For example we used to have many helper methods here which I replace with simple copy/paste. 
But in this specific case, I agree. Having something like:
```go
var err error
recordEndFunc := recordActivityStart(...)
defer recordEndFunc(err)
```
looks very natural for go developers. Similar to `ctx, cancel := context.WithCancel(ctx)`.

ctxpropagation/worker/main.go
```diff
@@ -3,24 +3,30 @@ package main
 import (
 	"log"
 
-	"github.com/opentracing/opentracing-go"
+	"github.com/temporalio/samples-go/ctxpropagation"
```
alexshtin, line 6: 3 groups separated by empty line, please:
```go
"std imports"

"3rd party imports"

"this project local imports"
```

alexshtin (approved):

ctxpropagation/starter/main.go
cretz: True. In fact, you'll note internally this can never return an error right now and I just added error capability for future proofing if we add options needing more advanced validation. So it's a perfect candidate for panicking.

Issue opened at https://github.com/temporalio/sdk-go/issues/664

metrics/metrics.go
cretz: My goal here was to keep as much the same as reasonable to update the SDK. I think a general question of improving this sample or any other is valid, but separate.

ctxpropagation/worker/main.go
cretz: :+1: Will change. I didn't realize this was by intention. I also haven't been requiring others do this in their contributions.

cretz: Fixed

PR #161, opened 2021-12-15
Request/response handling sample

### LINKED ISSUE #159: Add sample for bidirectional req/resp
Add a sample that shows the following methods of bidirectional communication:

* Signal + query poll
  * Local side: Send signal with ID to do something
  * Remote side: Store something with said ID and have query handler return said something when queried with ID
  * Local side: Query regularly to get result
* Signal + remote activity schedule
  * Local side: Have centralized notifier that has an activity to notify a waiting listener
  * Local side: Create ID-based listener on notifier and then send signal with task queue and ID
  * Remote side: Do something and start activity on given task queue and ID

cretz:
## What was changed

Demonstrates the primitive query-poll method and robust activity-callback method of handling responses from workflows.

## Why?

Requested in #159 to show how to use activities to push results of a request.

## Checklist

1. Closes #159


mmcshane (approved):
My only comment is that the discussion of draining and continue-as-new in the first section of the README distracts from the main thrust of the sample. You might consider moving it into a less emphasized "details" (or something like that) section so that the focus remains on Query/Signal/Response.

cretz:
@mmcshane - Moved continue-as-new explanation to its own section

samarabbas:
Thanks @cretz for creating this.  Looks good to me.

samarabbas (approved):

reqresp/README.md
```diff
@@ -0,0 +1,73 @@
+# Request/Response Sample
+
+This sample demonstrates how to send a request and get a response from a Temporal workflow. Two approaches to providing
+the caller a response are in the example:
+
+* Query poll - Caller can run queries polling for response (the slow/inefficient way)
+* Response activity - Caller can provide activity information on where to send the response to (the more robust way)
+
```
mfateev, line 8: I think _more robust_ is not really correct. Polling is robust. It just adds latency and can be less efficient.

reqresp/README.md
```diff
@@ -0,0 +1,73 @@
...
+When a query is executed in Temporal, it must go to the worker, invoke the query handler (potentially replaying if the
+workflow is not already present/cached), and return the query back. Therefore, in order to do a query-based
+request/response, one must signal the workflow and continuously query the workflow for the response. The query can't be
+executed too frequently as this affects server and worker resources. This is a method that, while simple, is much
+slower.
+
+Alternatively, a Temporal workflow can execute activities on different task queues. Therefore, a "response" task queue
+can be setup to handle activities executed from the other workflow. This means the response can be pushed. By providing
+which activity and task queue the response should be sent to inside the request, the local requester can just be
+notified when the local activity is executed.
```
mfateev, line 52: I'm confused. This explanation mentions both "task queues" and "local activities" which are not really compatible.

reqresp/README.md
```diff
@@ -0,0 +1,73 @@
...
+although the sample just shows it used ephemerally as part of the `request` CLI.
+
+### Explanation of continue-as-new
+
+Workflows cannot have infinitely-sized history and when the event count grows too large, `ContinueAsNew` can be returned
+to start a new one atomically. However, in order not to lose any data, signals must be drained and any other futures
+that need to be reacted to must be completed first. This means there must be a period where there are no signals to
+drain and no futures to wait on. If signals come in faster than processed or futures wait so long there is no idle
+period, `ContinueAsNew` will never happen and the history will grow unchecked until the extreme limit is reached at the
+server causing workflow termination.
```
mfateev, line 64: I don't believe this is the case. I think there is a hard limit on a number of signals a workflow can receive. After that the new signals are rejected, but the workflow is not terminated.

reqresp/requester.go
```diff
@@ -0,0 +1,218 @@
...
+	// Send request
+	return r.requestUppercase(ctx, &Request{ID: uuid.New(), Input: str})
+}
+
+func (r *requesterQueryResponses) Close() {
+	r.ticker.Stop()
+	r.pollCancel()
+}
+
+func (r *requesterQueryResponses) poll() {
```
mfateev, line 179: Why do it here instead of just after the Signal call? That would be much more simple.

reqresp/workflow.go
```diff
@@ -0,0 +1,195 @@
...
+		responses:                   make(map[string]*Response, 500),
+
+		// We're gonna just support 10 second timeout and 3 retries on response
+		// callback activities. WARNING: The timeout and retry affect how long this
+		// workflow stays open and may prevent it from performing its
+		// continue-as-new until timeout occurs and/or retries are finished.
+		responseActivityOptions: workflow.ActivityOptions{
+			// We use schedule-to-close because if the requester side is not present,
+			// this may hang otherwise with just start-to-close.
+			ScheduleToCloseTimeout: 10 * time.Second,
```
mfateev, line 99: Consider setting a short scheduleToStart to handle requester not present situation.

reqresp/workflow.go
```diff
@@ -0,0 +1,195 @@
...
+	// Listen for new requests
+	var requestCount int
+	selector.AddReceive(u.requestCh, func(c workflow.ReceiveChannel, more bool) {
+		requestCount++
+		var req Request
+		c.Receive(u, &req)
+		u.addExecuteActivityFuture(&req, selector)
+	})
+
+	// Continually select until there are too many requests and no pending selects
```
mfateev, line 129: I would expand the "no pending selects" comment. The "draining" is mentioned in some other comment, but this is exactly what guarantees it.

reqresp/README.md
cretz: Will update

reqresp/README.md
cretz: Sorry, I meant the word "local" as in "on this worker". I will use another term.

reqresp/README.md
cretz: I just assumed that from https://docs.temporal.io/docs/go/workflows/#large-event-histories. I was not aware of a separate signal limit. I will remove this detail anyways because the reader should not be concerned with what happens in this extreme case we're trying to avoid anyways.

reqresp/requester.go
cretz: So that the requester can be used for many calls. Consider having one "requester" for an entire microservice that handles many requests. By polling in this way, you can get multiple responses as a time. Note how the query accepts a collection of IDs we are looking for responses to.

While this sample's CLI is ephemeral, the requester is built for long-term thread-safe use.

I could also call poll immediately after signal and handle debouncing if many happened at a time, but this was the simplest logic while keeping the response multiplexing functionality. If it's worth it, I can send the query right away per request.

reqresp/workflow.go
cretz: I figured this setting was enough for that too since the value is so short and I expect the start-to-close to be very short. But I can set a schedule-to-start just for this demonstration just in case people repurposed the pattern with heavy response-handling activities.

reqresp/workflow.go
cretz: Will do. In fact, I've found the single-central-selector pattern to be so valuable, I am wondering if we should document this "reactor pattern" and encourage its use more (reminds me of actors).

reqresp/README.md
cretz: Updated

reqresp/README.md
cretz: Updated to avoid the word "local"

reqresp/README.md
cretz: Updated to avoid details about what happens on uncapped signal count

reqresp/workflow.go
cretz: Added a `ScheduleToStart` that is only 5 seconds. I kept `ScheduleToClose` because I care about a completion time limit too.

reqresp/workflow.go
cretz: Expanded.

cretz:
After discussion, we are going to change this in two ways:

* Separate samples into `reqrespquery` and `reqrespactivity` (and lose the abstraction over the two)
* Make requester CLI continually request on an interval until stopped and have README suggest running two at once to demonstrate task queue routing

cretz:
I am closing this in favor of a new PR since the changes will be too drastic to see here.

PR #164, opened 2022-01-04
JWT auth sample

### LINKED ISSUE #160: Add sample for auth token
Show how to add auth tokens via header providers and/or gRPC dial options

cretz:
## What was changed

Added a sample demonstrating manual/programmatic use of JWT with the server.

## Why?

Re #160, there is nothing out there showing how to use JWT w/ the server from the SDK.

## Checklist

1. Closes #160


sergeybykov:
One typical mistake we see people making is loading certs/tokens once during initialization of a long-running process. Everything works fine for awhile until the cert/token expires. Would it be helpful in the worker code to include logic to dynamically load or refresh the token? It could be as simple as a ticker. A little bit of extra complexity, but hopefully helping people avoid the frequent mistake.

serverjwtauth/key/main.go
```diff
@@ -0,0 +1,104 @@
...
+		Key:         key,
+		KeyID:       jwk.KeyID,
+		Permissions: []string{"system:admin"},
+		Expiration:  1 * time.Hour,
+	}
+	token, err := config.GenToken()
+	if err != nil {
+		return err
+	}
+	fmt.Printf("TEMPORAL_CLI_AUTHORIZATION_TOKEN=Bearer %v", token)
```
bergundy, line 102: You should probably quote the output here

serverjwtauth/worker/main.go
```diff
@@ -0,0 +1,45 @@
...
+				},
+			},
+		},
+	})
+	if err != nil {
+		log.Fatalln("Unable to create client", err)
+	}
+	defer c.Close()
+
+	w := worker.New(c, "hello-world", worker.Options{})
```
bergundy, line 36: nit: change the name of the task queue

serverjwtauth/README.md
```diff
@@ -0,0 +1,113 @@
...
+When starting the server using the docker container, the following excerpt will appear in the logs:
+
+    "msg":"uncategorized error","operation":"RegisterNamespace","wf-namespace":"default","error":"Request unauthorized."
+
+This is because by default the docker container attempts to register the `default` namespace, but cannot because we have
+restricted access to the server to only authorized uses.
+
+To create this manually, we will use `tctl`. According to
+[the tctl docs](https://docs.temporal.io/docs/devtools/tctl/#securing-tctl) we can use the `tctl-authorization-plugin`
+binary. If using docker, this binary is included with tctl otherwise, built it and put it on the `PATH`.
```
sergeybykov, line 69:
```suggestion
binary. If using docker, this binary is included with tctl. Otherwise, build it and put it on the `PATH`.
```

sergeybykov (approved):

serverjwtauth/README.md
robholland: Note that the homebrew install of `tctl` also installs this plugin. The only time someone would need to build and install the plugin is if they are running a tctl they have compiled themselves which seems a bit of an edge case for working from a sample.

robholland:
@sergeybykov The code already handles rotation. The header provider handles this in `GetHeaders` so the worker does not need to be aware of it.

robholland (approved):
Looks good, just need to change the task queue name 👍

serverjwtauth/key/main.go
cretz: I think it's best if the caller does. Quoting is specific to the shell it runs in (though probably blindly double quoting is fine).

serverjwtauth/README.md
cretz: I don't use mac and a lot of people don't use mac and sadly we don't provide executables for other platforms.

PR #166, opened 2022-01-05
Request/response handling samples

### LINKED ISSUE #159: Add sample for bidirectional req/resp

cretz:
## What was changed

Demonstrates the activity-callback and query-poll methods of handling responses from workflows.

## Why?

Requested in #159 to show how to use activities to push results of a request. Requested that the samples be split in the previous PR #161.

## Checklist

1. Closes #159
2. Successor to #161


reqrespactivity/requester.go
```diff
@@ -0,0 +1,130 @@
...
+		options:         options,
+		taskQueue:       "requester-" + uuid.New(),
+		pendingRequests: map[string]chan<- *Response{},
+	}
+	if r.options.ExistingWorker == nil {
+		r.options.ExistingWorker = worker.New(options.Client, r.taskQueue, worker.Options{})
+	}
+
+	// Start worker
+	r.options.ExistingWorker.RegisterActivity(r.responseActivity)
```
robholland, line 56: As far as I can tell would mean that the worker will poll on an increasing number of task queues. It's not clear to me if this is factored this way simply because of the constraints of making a sample or if this is a suggestion for a useable pattern. I appreciate the need for separated task queues so that the response lands in the same process that it was requested in but this does feel a bit odd.

robholland (approved):
Code wise looks ok to me. I feel like asynchronously completed activities is another viable pattern here when the requestor is itself a workflow, but I don't know what use case is behind these samples.

reqrespactivity/requester.go
cretz: Yes, there is a 1:1 correlation with workers/taskqueues and requesters. If you have 50 places sending request/response, you have 50 workers polling on separate task queues. For the most part an open long poll on each is mostly just a harmless HTTP connection (especially when I complete https://github.com/temporalio/sdk-go/issues/636 here in just a bit). Is it really any different than 50 client-side `ExecuteWorkflow` requests that all long poll on `Get` when awaiting response?

I definitely encourage long-term use of this "Requester", not per call. In fact, I would suggest only one per process.

I am open to alternatives of course.

cretz:
> I feel like asynchronously completed activities is another viable pattern here when the requestor is itself a workflow, but I don't know what use case is behind these samples.

@robholland - you can actually do cooler things when the requester is itself a workflow. Workflows talking to workflows get bidirectional signalling which obviates the need for almost all of these other approaches :-). In fact, this is how I implemented it in a toy code generator I made recently, see [here](https://github.com/cretz/temporal-sdk-go-advanced/blob/ccffaf6efa7e2ed1cb9f43bac23702cf831713ef/temporalproto/test/simplepb/simple_temporal.pb.go#L754-L813). (and I believe query responses or bidirectional signals or whatever is something being considered for native inclusion)

reqrespactivity/requester.go
robholland: The difference is that the pollers would stop long polling once they've got their answer. I may have missed it in the code but it looked to me that the activity is registered and never de-registered (is that even a thing?) which means it could add up to a pretty large number of pollers who will never see any traffic again (depending on request count, process lifetime etc etc).

cretz: Yes, I think there's some confusion here.

Say this requester should be reused for the all requests for the lifetime of the process. That means only one worker (so only one poller) and one activity registered to it. There is not a response activity per request. There is no need to unregister an activity, the activity is 1:1 with the worker.

So basically, if used properly, there is only one of these per process.

cretz: > The difference is that the pollers would stop long polling once they've got their answer.

I do not believe this is true. Activity pollers will continuously poll for the lifetime of the worker.

robholland: Indeed I was confused, I mis-read 'requester-uuid' as 'request-uuid' and had inferred it was one task queue per request/response pair, not per process. Makes sense. My statement about pollers stopping was referring to your mention of people calling StartWorkflow and then polling for the result. I believe they would generally stop polling once they got the result.

Sushisource (approved):

PR #170, opened 2022-01-18
Interceptor Sample

### LINKED ISSUE #150: Interceptor Samples
Once https://github.com/temporalio/proposals/pull/45 and https://github.com/temporalio/sdk-go/issues/529 are complete, there should be a sample for interceptors.

cretz:
## What was changed

Added simple interceptor example to show how to affect `GetLogger` with custom tags.

## Why?

There was no example showing how to create and configure interceptors. This specific one shows how to go all the way to outbound.

## Checklist

1. Closes #150
2. Closes https://github.com/temporalio/sdk-go/pull/668


bergundy (approved):
It would probably be more useful to have the custom logger be context aware and instead of adding statically defined tags, have it call a function that gets the workflow / activity context.

cretz:
@bergundy - Can you clarify a bit? The logger has a fixed set of methods it must implement and these do not take a context. You cannot obtain a context without it in the parameter.

cretz:
@bergundy - I have changed it to add workflow/activity start time as log tags

bergundy:
Nice! I've already approved before, merge at wiil

PR #176, opened 2022-02-22
Convert encryption example to remote data encoding.

robholland:
## What was changed
Updated sample to use remote data encoder rather than tctl plugin.

## Why?
The remote data encoding architecture is the preferred way to manage decryption in tctl/Temporal Web.

## Checklist

1. How was this tested:
Ran through the sample README and confirmed the steps behave as expected.


robholland:
Waiting on release of https://github.com/temporalio/sdk-go/pull/733

PR #179, opened 2022-03-11
Add grpc proxy example.

robholland:
## What was changed
Adds an example of building a GRPC proxy using the new primitives provided by the SDK.

## Why?
This allows people to centrally deploy codec for use by tctl and Temporal Web.

## Checklist

1. How was this tested:
The README was followed and confirmed that the commands behave as expected.


grpc-proxy/README.md
```diff
@@ -0,0 +1,22 @@
+### Steps to run this sample:
```
cretz, line 1: I know we haven't been doing it in many of our other examples, but can we get an overview here on what this demonstrates? E.g. "This sample shows how to compress data via a custom gRPC proxy using a custom codec".

grpc-proxy/data_converter.go
```diff
@@ -0,0 +1,67 @@
...
+func NewPayloadCodec() converter.PayloadCodec {
+	return &Codec{}
+}
+
+// Codec implements converter.PayloadEncoder for snappy compression.
+type Codec struct{}
+
+// Encode implements converter.PayloadCodec.Encode.
+func (e *Codec) Encode(payloads []*commonpb.Payload) ([]*commonpb.Payload, error) {
+	result := make([]*commonpb.Payload, len(payloads))
```
cretz, line 26: _Technically_ to get the best compression, the input would be wrapped in a single `commonpb.Payloads`, that would be encrypted and returned back as a single-item slice (and the inverse on `Decode`). But I don't necessarily think it's a requirement.

grpc-proxy/README.md
```diff
@@ -0,0 +1,22 @@
...
+1) You need a Temporal service running. See details in README.md
+2) Run the following command to start the worker
+```
+go run worker/main.go
+```
+3) Run the following command to start the example
+```
+go run starter/main.go
```
cretz, line 9: Does this not need the proxy to be running and therefore fail? Wouldn't we want the client and worker to connect to the proxy? I understand that we are showing a proxy created later, but I think it's clearer if we show that everything connects to that proxy instead of implementing data converters.

grpc-proxy/README.md
robholland: Added a brief overview.

grpc-proxy/data_converter.go
robholland: I don't think we should include such usage in samples yet as if for some reason it's copy/pasted into a system not using a proxy it could break Temporal Web for example, which never expects Payload count to change.

grpc-proxy/README.md
robholland: Updated to use proxy for worker/starter.

grpc-proxy/proxy-server/main.go
```diff
@@ -0,0 +1,71 @@
+package main
```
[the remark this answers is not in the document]
robholland: Discussed on Slack and agreed that a meaningful test would probably require a Temporal instance unlike the other samples in this repo. Given that, an integration test should be added to sdk-go rather than here.

cretz (approved):

PR #183, opened 2022-03-22
Support OIDC auth for remote codec server.

robholland:
## What was changed
Added support for OIDC auth to the remote codec server.

## Why?
This allows people to secure the remote codec server endpoint.

## Checklist

1. How was this tested:
Tested against http://github.com/temporalio/web/pull/445


robholland:
Making draft until Temporal Web and tctl remote decoding token support are released.

remote-codec/remote-codec-server/oidc.go
```diff
@@ -0,0 +1,88 @@
...
+	"strings"
+
+	"go.temporal.io/server/common/authorization"
+	"go.temporal.io/server/common/config"
+	"go.temporal.io/server/common/log/tag"
+)
+
+type Provider struct {
+	Issuer   string `json:"issuer"`
+	JWKS_URI string `json:"jwks_uri,omitempty"`
```
cretz, line 15: I know it seems weird to combine these, but still not bet to use an underscore here

remote-codec/remote-codec-server/oidc.go
```diff
@@ -0,0 +1,88 @@
+package main
```
cretz, line 1: Hrmm, can't think of an easy way necessarily, but would be nice if a real Go embedded oauth server could be used and a test existed. But none of that is required for now.

cretz (approved):

remote-codec/remote-codec-server/oidc.go
robholland: What would it be? `JWKSURI`?

PR #184, opened 2022-03-23
GRPC proxy OIDC

robholland:
## What was changed
Add support for OIDC authentication to the proxy.

## Why?
This is required to secure the GRPC proxy endpoints as you would Temporal itself.

## Checklist

1. How was this tested:


grpc-proxy/proxy-server/oidc.go
```diff
@@ -0,0 +1,62 @@
+package main
```
cretz, line 1: Ideally we would have tests confirm this works and will work into the future. But I understand if it's not something you want to do right now.

cretz (approved):

PR #197, opened 2022-05-02
Update to metrics sample

tsurdilo:
Signed-off-by: Tihomir Surdilovic <tihomir@temporal.io>

* adds how to add custom metric in workflow code
* adds how to set custom tags
* updates worker to publish metrics on port 9092 (default server metrics are exposed on port 9090 so this way it wont clash).

### DIFF
--- a/metrics/worker/main.go
+++ b/metrics/worker/main.go
@@ -18,7 +18,7 @@ func main() {
 	// The client and worker are heavyweight objects that should be created once per process.
 	c, err := client.NewClient(client.Options{
 		MetricsHandler: sdktally.NewMetricsHandler(newPrometheusScope(prometheus.Configuration{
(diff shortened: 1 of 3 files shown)


metrics/worker/main.go
```diff
@@ -55,6 +55,8 @@ func newPrometheusScope(c prometheus.Configuration) tally.Scope {
 		Separator:       prometheus.DefaultSeparator,
 		SanitizeOptions: &sanitizeOptions,
 		Prefix:          "temporal_samples",
+		Tags:            map[string]string{"my_key1": "my_value1",
+			                       "my_key2": "my_value2"},
```
cretz, line 59: You need to run `go fmt` on this file. All Go IDEs should do this automatically for you on save.

tsurdilo: @cretz done

PR #206, opened 2022-06-28
Added vscode remote containers config

### LINKED ISSUE #205: [Feature Request] Add support for codespaces/devcontainers for easy local development
### Is your feature request related to a problem? Please describe.

VScode has a cool feature called Remote Containers. It would be nice to leverage this so developers can spin up a temporal service right on their local machines for easy development. 

I'd like to take the same gitpod docker-compose.yml file and modify it to work with VScode remote containers.

### Additional context

I will submit a pull request for this very soon.

clintberry:
This allows easy local development with VScode Remote Container plugin (free)

## What was changed
Added devcontainer configuration for easy local development on VScode

## Why?
This allows you to run a dev temporal service locally for easy developing of temporal apps

## Checklist

1. Closes #205 

2. How was this tested:
- Install remote containers extension in VScode
- clone and checkout this repo and open in vs code
- Open vs code command terminal and search for action "Remote Containers: Rebuild and Reopen in container
- After it is finished spinning up, click on ports tab in vscode and click on "localhost:8080" to load the temporal web interface locally

3. Any docs updates needed?
We could add another section titled "Remote Container" on this page: https://docs.temporal.io/application-development-guide#gitpod

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ This repository contains several sample Workflow applications that demonstrate t
 ## How to use
 
 - You can run this in the browser with Gitpod: [![Gitpod ready-to-code](https://img.shields.io/badge/Gitpod-ready--to--code-908a85?logo=gitpod)](https://gitpod.io/#https://github.com/temporalio/samples-go/)
-- Or you can run Temporal Server locally using [VSCode Remote Containers](https://code.visualstudio.com/docs/remote/containers). (install Remote containers plugin in VScode and then open this repo and select "Run in container" when prompted)
+- Or you can run Temporal Server locally using [VSCode Remote Containers](https://code.visualstudio.com/docs/remote/containers). [![Open in Remote - Containers](https://img.shields.io/static/v1?label=Remote%20-%20Containers&message=Open&color=blue&logo=visualstudiocode)](https://vscode.dev/redirect?url=vscode://ms-vscode-remote.remote-containers/cloneInVolume?url=https://github.com/temporalio/samples-go) 
 - Lastly, you can run Temporal Server locally on your own (follow the [Quick install guide](https://docs.temporal.io/docs/server/quick-install)), then clone this repository
 
 The [helloworld](helloworld/README.md) sample is a good place to start.


cretz:
Can you clarify something about codespaces? It seems with GitPod, it's the user that pays and can upgrade and what not, but with codespaces, it is the repo/org? Also, it seems codespaces unlike GitPod does not have a free tier anyone can use, correct?

clintberry:
That is very true. Maybe I should have just left out that the same config can be used for codespaces. The real benefit for me here is that I can spin it up locally in a VScode devcontainer (just uses docker compose with some other niceties). Do you want me to change the issue/commit message?

cretz:
Yeah, let's leave off anything about codespaces. Feel free to update the README on how vscode devcontainer can be used (can be an additional bullet on how-to-use that links to vscode docs if needed). I will review the rest.

cretz:
Also, feel free to provide a `vscode://` link right on the repo so people can click it.

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,20 @@
+# See here for image contents: https://github.com/microsoft/vscode-dev-containers/tree/v0.231.6/containers/go/.devcontainer/base.Dockerfile
+
+# [Choice] Go version (use -bullseye variants on local arm64/Apple Silicon): 1, 1.16, 1.17, 1-bullseye, 1.16-bullseye, 1.17-bullseye, 1-buster, 1.16-buster, 1.17-buster
+ARG VARIANT="1.18-bullseye"
```
cretz, line 4: Will this work out of the box for a Windows or Linux x64 user out of the box or must they come in here and change this value? Or does `[Choice]` mean that the user will be presented with an option? If so, can the default just be "1.18" and also can "1.18" be added in the list?

.devcontainer/docker-compose.yml
```diff
@@ -0,0 +1,86 @@
+# this is a simple docker compose that has been tested to work with remote containers in VS code.
```
cretz, line 1: Sometimes sentences start with capitals, sometimes they don't

.devcontainer/docker-compose.yml
```diff
@@ -0,0 +1,86 @@
...
+# You can run these samples in the browser with no need to run Temporal Server locally
+# by using Codespaces on github. 
+# see https://github.com/temporalio/docker-compose for other reference docker-compose files
+version: '3.5'
+services:
+  ## uncomment if you need search attributes feature
+  ## dont forget to uncomment the ES related comments in the temporal service below
```
cretz, line 8: don't* (and should in general add sentence punctuation)

clintberry:
Since VScode remote containers isn't a hosted service like GitPod, a direct link wouldn't be useful because the person still needs to check out the code manually and open in VScode. Hope that makes sense.

cretz:
You can leave "codespaces" stuff in the PR btw. I have confirmed we do this in https://github.com/temporalio/samples-typescript/. Sorry about teh confusion.

> a direct link wouldn't be useful because the person still needs to check out the code manually and open in VScode.

I see https://github.com/microsoft/vscode-remote-try-go has a badge with a link like: `https://vscode.dev/redirect?url=vscode://ms-vscode-remote.remote-containers/cloneInVolume?url=https://github.com/microsoft/vscode-remote-try-go`. I wonder if something similar can happen here.

.devcontainer/docker-compose.yml
```diff
@@ -0,0 +1,86 @@
...
+      - DB=postgresql
+      - DB_PORT=5432
+      - POSTGRES_USER=temporal
+      - POSTGRES_PWD=temporal
+      - POSTGRES_SEEDS=postgresql
+      - DYNAMIC_CONFIG_FILE_PATH=config/dynamicconfig/development_es.yaml
+    #  - ENABLE_ES=true
+    #  - ES_SEEDS=elasticsearch
+    #  - ES_VERSION=v7
+    image: temporalio/auto-setup:1.13.4
```
cretz, line 62: This is an old version. Maybe `temporalio/auto-setup:latest` would be better (same with other Temporal images herein).

clintberry: 🤦 sorry about that. latest is definitely better.

.devcontainer/Dockerfile
clintberry: I can change it to "1" and it will always pull the latest version of the Go docker container (for version 1 that is). Bullseye works with windows and linux, but if you are an arm mac user I don't think "1" will work right now as they don't seem to have arm64 images for every release yet. I think the best bet for max compatibility and getting most recent Go version would be "1-bullseye"

clintberry:
I also wasn't aware this was already done for the Typescript samples. That repo uses a different container I didn't know about called temporal lite from data dog. Maybe I should switch to that same image since it makes the footprint a lot smaller locally.
https://github.com/temporalio/samples-typescript/blob/main/.devcontainer/docker-compose.yml

.devcontainer/Dockerfile
clintberry: yes, bullseye has both x64 and arm, but the other go images only have x64 still. non-arm people can use the -bullseye image without issue and it will pull the correct image for them automatically.

clintberry:
Let's keep it with the official images so they stay up to date.

clintberry:
The button works! so cool. I'm adding that to all our company repos just to make it easier for devs.

cretz (approved):
Approved. Thanks! Feel free to merge whenever

clintberry:
I don't have write access to this repo :-)

cretz:
Merged, thanks again!

PR #208, opened 2022-07-13
Use errors.Is instead of == for error comparison

alfa-alex:
This allows wrapped errors to be matched correctly.

## What was changed

Changed an error check that was previously performed via `err != temporal.ErrNoData` to use `errors.Is` instead.

## Why?
Using `errors.Is` is the preferred way since Go 1.13. `!errors.Is(err, temporal.ErrNoData)` returns `false` even if `temporal.ErrNoData` is wrapped inside `err` (and not literally equal).

## Checklist
The change is so minor that it should be possible to just carefully check the diff.

### DIFF
--- a/reqrespquery/requester.go
+++ b/reqrespquery/requester.go
@@ -2,6 +2,7 @@ package reqrespquery
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"log"
 	"time"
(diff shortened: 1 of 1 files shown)


alfa-alex:
... if you want to rely on it never being wrapped in the future.

Alright, sorry for the noise then!

PR #209, opened 2022-07-24
Allow ability to load client options from config

samarabbas:
## What was changed
- Added helper methods to load config for constructing client options
- Modified helloworld sample to use the newly added helper method

User can now control different configs through following environment variables
```
TEMPORAL_CONFIG_ROOT=".temporal"
TEMPORAL_ENVIRONMENT="local"
TEMPORAL_CONFIG_FILE  = "config.yaml"
```

Config File Format
```
host: "localhost:7233"
namespace: "default"
caFile: ""
certFile: ""
keyFile: ""
caData: ""
certData: ""
keyData: ""
```

## Why?

Allowing ability to easily load client options from config enables user to run
samples against any environment including Temporal Cloud

## Checklist

1. Closes 

2. How was this tested:

Manually ran this against both local docker-compose setup and Temporal Cloud

3. Any docs updates needed?

### DIFF
--- a/common/connectivity.go
+++ b/common/connectivity.go
@@ -1,0 +1,134 @@
+package common
+
+import (
+	"crypto/tls"
+	"crypto/x509"
+	"encoding/base64"
+	"errors"
+	"fmt"
+	"io/ioutil"
+	"net"
+	"os"
+	"path/filepath"
+
+	"go.temporal.io/sdk/client"
+)
+
+func GetClientOptions() client.Options {
+	configHome := getEnvOrDefaultString(EnvKeyConfigHome, EnvDefaultConfigHome)
+	configEnv := getEnvOrDefaultString(EnvKeyEnvironment, EnvDefaultEnvironment)
+
+	configRoot := filepath.Join(configHome, configEnv)
+	// Use Home Directory to contruct the configRoot if it is specified as relative path
+	if !filepath.IsAbs(configRoot) {
+		homeDir, err := os.UserHomeDir()
+		if err != nil {
+			panic(err)
+		}
+		configRoot = filepath.Join(homeDir, configRoot)
+	}
+
+	if _, err := os.Stat(configRoot); err != nil {
+		// Config root does not exist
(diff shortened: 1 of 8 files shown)


common/connectivity.go
```diff
@@ -0,0 +1,134 @@
...
+
+	if len(caBytes) > 0 {
+		caPool = x509.NewCertPool()
+		if !caPool.AppendCertsFromPEM(caBytes) {
+			return nil, errors.New("unknown failure constructing cert pool for ca")
+		}
+	}
+
+	// If we are given arguments to verify either server or client, configure TLS
+	if caPool != nil || cert != nil {
```
cretz, line 98: Many people want TLS and don't need to configure mTLS or a custom server CA. Often people host Temporal behind a server cert obtained from a valid CA and no mTLS.

common/connectivity.go
```diff
@@ -0,0 +1,134 @@
...
+		tlsConfig := &tls.Config{
+			InsecureSkipVerify: !config.EnableHostVerification,
+			ServerName:         host,
```
cretz, line 101: This is already assumed by normal TLS IIRC (they use the hostname of the dial) and usually you don't want SNI override unless you really need to change it.

common/loader.go
```diff
@@ -0,0 +1,50 @@
...
+	EnvDefaultConfigFile  = "config.yaml"
+)
+
+func loadConfig(configRoot string, config interface{}) error {
+	configFile := getEnvOrDefaultString(EnvKeyConfigFile, EnvDefaultConfigFile)
+
+	configPath := filepath.Join(configRoot, configFile)
+	fmt.Printf("Loading config from: %v\n", configPath)
+
+	data, err := ioutil.ReadFile(configPath)
```
cretz, line 30: `ioutil` is deprecated, prefer `os.ReadFile`

cretz:
Closing for now pending larger discussion about making all of our samples non-localhost for use with Temporal cloud and other non-local servers.

PR #213, opened 2022-08-11
Multi-namespace client POC

cretz:
## What was changed

Added POC for a multi-namespace client as requested internally. The client is very naive and the hashing is _intentionally not consistent_ if the state of the namespaces change.

Opening as draft for review only. Probably will never be merged.


multinsclient/client.go
```diff
@@ -0,0 +1,177 @@
...
+	if len(c.enabledNamespaces) == 0 {
+		return nil, fmt.Errorf("no enabled namespaces")
+	}
+
+	success = true
+	return c, nil
+}
+
+// NamespaceFor selects a namespace and client for the given workflow ID. The
+// resulting namespace is always the same for the same workflow ID.
```
yiminc, line 115: > The resulting namespace is always the same for the same workflow ID

This is not true if onlyEnabled = true and there is disabled namespace.

multinsclient/client.go
```diff
@@ -0,0 +1,177 @@
...
+//
+// If onlyEnabled is true and the namespace that would be selected is not
+// enabled, a hash is done only across the enabled namespace. Therefore, while
+// the same namespace is always selected even after Reset for the same workflow
+// ID when onlyEnabled is false, when onlyEnabled is true, the selected
+// namespace may be different for each Reset.
```
yiminc, line 121: This is not very clear, and may be simplified. 
If onlyEnabled is false, a namespace will be picked among all namespaces based on deterministic hash regardless if picked namespace is disabled or not. If onlyEnabled is true, and the namespace picked is disabled, a second pick will be done among enabled namespaces.

yiminc (commented):
We need a wrapper for ListWorkflow API as well.

multinsclient/client.go
```diff
@@ -0,0 +1,177 @@
...
+	Namespaces []Namespace
+
+	// Hasher that chooses a number from 0 (inclusive) to max (exclusive) for the
+	// given workflow ID. If unset, DefaultHasher is used.
+	Hasher func(workflowID string, max int) int
+
+	// If true, enabled namespaces will create clients that won't connect until
+	// first used. Otherwise, enabled namespaces will have their clients eagerly
+	// attempt to connect to the server when the client is created.
+	ConnectLazy bool
```
sergeybykov, line 43: Do we really need this? I would take it out for simplicity.

multinsclient/client.go
```diff
@@ -0,0 +1,177 @@
...
+// resulting namespace is always the same for the same workflow ID.
+//
+// If onlyEnabled is true and the namespace that would be selected is not
+// enabled, a hash is done only across the enabled namespace. Therefore, while
+// the same namespace is always selected even after Reset for the same workflow
+// ID when onlyEnabled is false, when onlyEnabled is true, the selected
+// namespace may be different for each Reset.
+//
+// The resulting client is only non-nil if the namespace is enabled.
+func (c *Client) NamespaceFor(workflowID string, onlyEnabled bool) (namespace string, client client.Client) {
```
sergeybykov, line 124: Do we actually need `onlyEnabled`? If there's no scenario where we expect it to be passed as `false` , we can simplify by taking it out and always falling back to enabled in case the primary match is not.

multinsclient/client.go
```diff
@@ -0,0 +1,177 @@
...
+	return client
+}
+
+// Reset creates a new client from the given options. The namespaces must be the
+// same count with the same names in the same order. If closeCurrent is true,
+// this will close the current client before returning a new client on success.
+//
+// Reset is used for creating a new client if which namespaces are enabled
+// changes. This does close and reconnect all client connections.
+func (c *Client) Reset(options Options, closeCurrent bool) (*Client, error) {
```
sergeybykov, line 148: `Reset` sounds a bit confusing to me, and it behaves more like clone or fork. Could we instead turn it into `Update` or `UpdateConfig` that would atomically switch to the new configuration internally and close old clients always?

cretz:
> We need a wrapper for ListWorkflow API as well.

Why can't they just `client.ClientFor("my-namespace").ListWorkflow(...` just like they would `client.ListWorkflow(...`? Why can't they type `ClientFor("my-namespace")`?

multinsclient/client.go
cretz: I don't know whether they use lazy connectivity or eager connectivity today.

multinsclient/client.go
```diff
@@ -0,0 +1,177 @@
...
+	// If true, enabled namespaces will create clients that won't connect until
+	// first used. Otherwise, enabled namespaces will have their clients eagerly
+	// attempt to connect to the server when the client is created.
+	ConnectLazy bool
+}
+
+// DefaultHasher uses FNV-1a hashing.
+func DefaultHasher(workflowID string, max int) int {
+	h := fnv.New32a()
+	h.Write([]byte(workflowID))
```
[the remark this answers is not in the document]
cretz: Yes, _very_ lightweight: https://github.com/golang/go/blob/43456202a1e55da55666fac9d56ace7654a65b64/src/hash/fnv/fnv.go#L48-L53

multinsclient/client.go
cretz: It's actually always true for _this_ client which has an immutable set of namespaces. I explain below how, after Reset, this could be different.

multinsclient/client.go
cretz: I think my first sentence is simpler than your two and that my second sentence discussing Reset is important and doesn't need to be removed. But if y'all think otherwise, ok.

multinsclient/client.go
cretz: If y'all are too concerned about having this available, I can make this method unexported. It's not like this is some complicated library I need to hide things in. If I were using this library, I might to know which namespace _would_ have been picked if all were enabled.

multinsclient/client.go
cretz: `Update` sounds like it mutates the receiver. Can you think of another name?

multinsclient/client.go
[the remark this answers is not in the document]
yiminc: 👍

multinsclient/client.go
yiminc: Yes, make this method unexpected, and have another one that exported that call into this one with the flag always set to true.

multinsclient/client.go
yiminc: I think we don't need this Reset method. When the config changed, just call the New() method to get new one.

yiminc:
> > We need a wrapper for ListWorkflow API as well.
> 
> Why can't they just `client.ClientFor("my-namespace").ListWorkflow(...` just like they would `client.ListWorkflow(...`? Why can't they type `ClientFor("my-namespace")`?

The idea is to show how this multiple namespaces can be hide behind this library. The ListWorkflow wrapper would return a wrapped next page token, which would contain the raw next page token for "current namespace" and index of current namespace. The wrapper of ListWorkflow will extract the wrapped page token and page through current namespace and then move on to next namespace (by increase current namespace index in wrapped next page token). But from consumer of this library, it does not see the multiple namespaces, and it just simply page through using the wrapped next page token.

multinsclient/client.go
cretz: Why? Why can't I as a user get what the namespace would be if they were all enabled? There is no benefit in limiting the user here.

multinsclient/client.go
cretz: Sure we don't need it. In fact, most of this PR is a "don't need" because it's so trivial. I think this method deserves to be here to understand the semantics of "reset".

cretz:
Added wrapper calls including `ListWorkflow`

cretz:
Closing PR, it has served its purpose.

PR #214, opened 2022-08-11
testing snipsync

rachfop:
## What was changed
Adds SnipSync. 
## Why?

Per this [plan adding](url) snip sync to code modules.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/helloworldmtls/helloworld.go
+++ b/helloworldmtls/helloworld.go
@@ -58,7 +58,7 @@ func ParseClientOptionFlags(args []string) (client.Options, error) {
 	} else if *clientCert == "" || *clientKey == "" {
 		return client.Options{}, fmt.Errorf("-client-cert and -client-key are required")
 	}
-	// @@@SNIPSTART go-mtls-configuration
+
 	// Load client cert
 	cert, err := tls.LoadX509KeyPair(*clientCert, *clientKey)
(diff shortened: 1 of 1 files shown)


helloworldmtls/helloworld.go
```diff
@@ -76,6 +76,7 @@ func ParseClientOptionFlags(args []string) (client.Options, error) {
 			return client.Options{}, fmt.Errorf("server CA PEM file invalid")
 		}
 	}
+	// @@@SNIPEND
```
bergundy, line 79: You should include the client options too

cretz, line 79: I wonder if we really need the custom server CA cert stuff. I did it here because it makes sense for self-hosted Temporal servers, but for cloud you probably won't need it. I believe our certs are signed by legit root CAs.

cretz: You don't even need `serverCAPool` except if you're hosting the server using a self-signed certificate (not to be confused with a self-signed client certificate). But I'll let @bergundy weigh in on whether we want to document how to do server CA override.

bergundy: I'd avoid this, I expect most of our users to not have to set this and work with proper CAs.

PR #216, opened 2022-09-12
Add upsert memo sample

rodrigozhou:
## What was changed

Create a sample demonstrating the usage of memo and upserting memo.

## Why?

N/A


memo/memo_workflow.go
```diff
@@ -0,0 +1,144 @@
...
+	workflowpb "go.temporal.io/api/workflow/v1"
+	"go.temporal.io/sdk/activity"
+	"go.temporal.io/sdk/client"
+	"go.temporal.io/sdk/converter"
+	"go.temporal.io/sdk/log"
+	"go.temporal.io/sdk/workflow"
+)
+
+/*
+ * This sample shows how to use memo. (Note this feature only work with Elasticsearch)
```
cretz, line 20: Why does it only work with Elasticsearch?

memo/memo_workflow.go
```diff
@@ -0,0 +1,144 @@
...
+	}
+
+	// Print current memo with modifications above.
+	info = workflow.GetInfo(ctx)
+	err = printMemo(info.Memo, logger)
+	if err != nil {
+		return err
+	}
+
+	// Now send commands to the server and let Elasticsearch update the index.
```
cretz, line 72: Unrelated to Elasticsearch. Also, I think it may be confusing to users that they have to sleep to have something sent to the server. I think that information can be left off (clearly this is a copy/paste from the old `searchattributes` example, but it may not be the best).

Also, we encourage people to ignore errors.

memo/memo_workflow.go
rodrigozhou: It was copy-paste, just updated PR to fix those comments.

memo/memo_workflow.go
rodrigozhou: Comments are fixed.

cretz (approved):

bergundy (commented):
I'd show that upsert doesn't overwrite fields that were not set in the `UpsertMemo` call and that setting to `nil` deletes a memo field.

PR #219, opened 2022-09-15
ApplicationError

finnigantime:


helloworld/helloworld.go
```diff
@@ -36,5 +77,7 @@ func Workflow(ctx workflow.Context, name string) (string, error) {
 func Activity(ctx context.Context, name string) (string, error) {
 	logger := activity.GetLogger(ctx)
 	logger.Info("Activity", "name", name)
-	return "Hello " + name + "!", nil
+
+	// return "Hello " + name + "!", nil
+	return "", temporal.NewApplicationError("some err", "NonRetryableError", NewNonRetryableErrorWithCode("foo msg", ErrorCodeFoo))
```
finnigantime, line 82: here we explicitly wrap as a `temporal.ApplicationError`

helloworld/starter/main.go
```diff
@@ -33,7 +35,25 @@ func main() {
...
+			log.Println("Cannot convert to NonRetryableError")
+		}
+
+		var appErr *temporal.ApplicationError
+		if errors.As(err, &appErr) {
+			log.Println("ApplicationError", appErr)
+			log.Println("HasDetails", appErr.HasDetails())
+			err = appErr.Details(&nonRetryErr)
+			log.Println("conversion err?", err)
+			log.Println("NonRetryableError", nonRetryErr.Err, *nonRetryErr.Code)
```
finnigantime, line 53: this works! we get out the error code

```
2022/09/15 16:35:29 Cannot convert to NonRetryableError
2022/09/15 16:35:29 ApplicationError some err (type: NonRetryableError, retryable: true)
2022/09/15 16:35:29 HasDetails true
2022/09/15 16:35:29 conversion err? <nil>
2022/09/15 16:35:29 NonRetryableError foo msg foo-code
2022/09/15 16:35:29 Workflow result:
```

PR #221, opened 2022-10-12
Add TLS related flags to helloworldmtls example

annismckenzie:
## What was changed

I added two options to the `helloworldmtls` example to be used when testing a Temporal Cluster that's secured with mTLS.

## Why?
In order to use the `helloworldmtls` example to test a Temporal Cluster that is secured via mTLS and uses self-signed certificates, either a) the server name contained in the server's certificate has to be passed or b) TLS verification needs to be disabled. I added both options.

## Checklist
1. Added `-server-name` and `-insecure-skip-verify` options to the `helloworldmtls` example.
2. Documented the change in the example's `README.md` file.

3. How was this tested:
Manually against our test cluster using the new options.

4. Any docs updates needed?
The additional options are documented in the example's `README.md` file.

### DIFF
--- a/helloworldmtls/README.md
+++ b/helloworldmtls/README.md
@@ -8,3 +8,7 @@ go run ./helloworldmtls/worker -target-host my.namespace.tmprl.cloud:7233 -names
 ```
 go run ./helloworldmtls/starter -target-host my.namespace.tmprl.cloud:7233 -namespace my.namespace -client-cert path/to/cert.pem -client-key path/to/key.pem
 ```
+
+If the server uses self-signed certificates, pass one of the following two options when starting the worker or the example above:
+1. `-server-name` and provide the common name contained in the self-signed server certificate
+2. `-insecure-skip-verify` which disables certificate and host name validation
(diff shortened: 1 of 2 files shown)


helloworldmtls/README.md
```diff
@@ -8,3 +8,7 @@ go run ./helloworldmtls/worker -target-host my.namespace.tmprl.cloud:7233 -names
 ```
 go run ./helloworldmtls/starter -target-host my.namespace.tmprl.cloud:7233 -namespace my.namespace -client-cert path/to/cert.pem -client-key path/to/key.pem
 ```
+
+If the server uses self-signed certificates, pass one of the following two options when starting the worker or the example above:
```
cretz, line 12:
```suggestion
If the server uses self-signed certificates and does not have the SAN set to the actual host, pass one of the following two options when starting the worker or the example above:
```

Just having self-signed is not a problem. The problem is they don't have the host you are trying to connect to in the SAN of the certificate. In theory we should discourage SNI override and skip-verify, but the options are fine to have in this sample.

annismckenzie: Thank you for the suggestion – you're absolutely right. 🎉

annismckenzie:
@cretz I rebased this onto main and was wondering if there's anything missing to get this merged? I'd like to update our own documentation and it would be nice not to carry around the fork. :)

cretz (approved):
Looks good. Running CI and then will merge.

annismckenzie:
Thank you! 🙇

PR #225, opened 2022-10-23
feat: add cancel-in-progress sample

StarpTech:
## What was changed

This example demonstrates how to implement a workflow that ensures that only one run of the child workflow is executed at a time. Subsequent runs will cancel the running child workflow and re-run it with the latest sent options through `SignalWithStartWorkflow`.

Those semantics are useful, especially when implementing a CI pipeline. New commits during the workflow execution should short-circuit the child workflow and only build the most recent commit.

## Why?

At [WunderGraph](https://github.com/wundergraph), we use temporal to build our CI and CD platform. The solution has been implemented in collaboration with @mfateev I promised him to make a PR :smiley: 

## Checklist

1. Closes 

2. How was this tested:

Manually, through the sample with a local temporal cluster.

3. Any docs updates needed?

I'm not familiar with the template boundaries `@@@SNIPSTART` in the samples. Please tell me the instructions.

### DIFF
--- a/cancel-in-progress/child_workflow.go
+++ b/cancel-in-progress/child_workflow.go
@@ -1,0 +1,19 @@
+package cancel_in_progress
+
+import (
+	"go.temporal.io/sdk/workflow"
+	"time"
+)
+
+// SampleChildWorkflow is a Workflow Definition
+func SampleChildWorkflow(ctx workflow.Context, name string) (string, error) {
+	logger := workflow.GetLogger(ctx)
+
+	// Simulate some long running processing.
+	_ = workflow.Sleep(ctx, time.Second*3)
+
+	greeting := "Hello " + name + "!"
+	logger.Info("Child workflow execution: " + greeting)
+
+	return greeting, nil
+}
--- a/cancel-in-progress/parent_workflow.go
+++ b/cancel-in-progress/parent_workflow.go
@@ -1,0 +1,106 @@
+package cancel_in_progress
+
+import (
+	"errors"
+	"go.temporal.io/sdk/log"
+	"go.temporal.io/sdk/workflow"
+)
+
+const ParentWorkflowSignalName = "parent-workflow-signal"
+
+func SampleParentWorkflow(ctx workflow.Context) (result string, err error) {
+	logger := workflow.GetLogger(ctx)
+
+	var message string
+
+	reBuildSignalChan := workflow.GetSignalChannel(ctx, ParentWorkflowSignalName)
+
+	// This will not block because the workflow is started with a signal transitionally.
+	reBuildSignalChan.Receive(ctx, &message)
+
+	cwo := workflow.ChildWorkflowOptions{
+		WaitForCancellation: true,
+	}
+	ctx = workflow.WithChildOptions(ctx, cwo)
+
+	var isProcessingDone = false
+
+	for !isProcessingDone {
+		ctx, cancelHandler := workflow.WithCancel(ctx)
+
+		// it is important re-execute the child workflow after every signal
+		// because we might have cancelled the previous execution
+		childWorkflowFuture := workflow.ExecuteChildWorkflow(ctx, SampleChildWorkflow, message)
+
+		selector := workflow.NewSelector(ctx)
+
+		selector.AddFuture(childWorkflowFuture, func(f workflow.Future) {
+			err = f.Get(ctx, &result)
+
+			// we don't want to end the parent workflow when child workflow is cancelled
+			if errors.Is(ctx.Err(), workflow.ErrCanceled) {
+				logger.Info("Child execution cancelled.")
+				return
+			}
+
+			if err != nil {
+				logger.Error("Child execution failure.", "Error", err)
+			}
+
+			// Child workflow completed. Let's end the parent workflow.
+			isProcessingDone = true
+		})
+
+		selector.AddReceive(reBuildSignalChan, func(c workflow.ReceiveChannel, more bool) {
+			logger.Info("Received signal.", "Message", message)
+
+			// cancel the child workflow as fast as possible after we received a new signal
+			// if a child workflow execution is in progress we cancel everything and start a new child workflow (for loop)
+			cancelHandler()
+
+			// drain the channel to get the latest signal
+			message = GetLatestMessageFromChannel(logger, reBuildSignalChan)
+		})
+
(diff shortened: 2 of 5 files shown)


cretz:
I'm not sure I understand this exactly, but there may be better ways to achieve what you are doing.

To confirm, this example just starts a child workflow when the parent workflow starts and then waits for a signal to cancel the child workflow and start a new one? If I send 3 signals you start 3 children without waiting for any of the children to complete?

Some notes:

* This could easily start dozens of child workflows and you'd never know if one errored. As a general purpose example, you may want to ensure only one child runs at a time.
* You can completely miss child workflow failure if signal comes in at same time which is not safe for a general purpose example.
* You're checking the parent context to see whether the child is cancelled.
* For a general sample, it may not make sense to drain the signal channel. If I send 2 signals as a caller, it may be strange that one was ignored.

Since this is essentially restart-child-on-signal, not sure it deserves an entire sample. But if so I think it could be simplified and clarified a bit. You might also want a test or two. For instance, make a test where you replicate the condition where the `"Child execution cancelled."` log appears.

StarpTech:
Hi, @cretz it would help a lot if you could point to the relevant places for your argumentation.

> You're checking the parent context to see whether the child is canceled.

I don't think so. The cancellable context is created here https://github.com/temporalio/samples-go/pull/225/files#diff-e7190ffa6a984a8e3be3db5ad97758314062293387073e586d442ddd9d664b81R29

> For a general sample, it may not make sense to drain the signal channel. If I send 2 signals as a caller, it may be strange that one was ignored.

That's the idea of this sample. I'm only interested in the last signal.

StarpTech:
> To confirm, this example just starts a child workflow when the parent workflow starts and then waits for a signal to cancel the child workflow and start a new one? If I send 3 signals you start 3 children without waiting for any of the children to complete?

No, I start a workflow with a signal. When another signal is sent during the execution of the child workflow the workflow is canceled and a new one is started with the options from the last signal.

cancel-in-progress/child_workflow.go
```diff
@@ -0,0 +1,19 @@
+package cancel_in_progress
+
+import (
+	"go.temporal.io/sdk/workflow"
+	"time"
```
cretz, line 5: I forget whether it's `go fmt` or `goimports` but may need to run them both so that stdlib imports are separate from others

cancel-in-progress/child_workflow.go
```diff
@@ -0,0 +1,19 @@
...
+)
+
+// SampleChildWorkflow is a Workflow Definition
+func SampleChildWorkflow(ctx workflow.Context, name string) (string, error) {
+	logger := workflow.GetLogger(ctx)
+
+	// Simulate some long running processing.
+	_ = workflow.Sleep(ctx, time.Second*3)
```
cretz, line 13: If you swallow this error this cannot get cancelled. It is important for our samples to show best practices of always responding to errors (if we haven't in other samples we should fix that)

cancel-in-progress/parent_workflow.go
```diff
@@ -0,0 +1,109 @@
+package cancel_in_progress
+
+import (
+	"errors"
+	"go.temporal.io/sdk/log"
+	"go.temporal.io/sdk/workflow"
+)
+
+const ParentWorkflowSignalName = "parent-workflow-signal"
```
cretz, line 9: IMO the signal name should reference what it does rather than who it's to, e.g. `const RestartChildSignalName = "restart-child-signal"`

cancel-in-progress/parent_workflow.go
```diff
@@ -0,0 +1,109 @@
...
+
+func SampleParentWorkflow(ctx workflow.Context) (result string, err error) {
+	logger := workflow.GetLogger(ctx)
+
+	var message string
+
+	reBuildSignalChan := workflow.GetSignalChannel(ctx, ParentWorkflowSignalName)
```
cretz, line 16: What do you mean here by "build"?

cancel-in-progress/parent_workflow.go
```diff
@@ -0,0 +1,109 @@
...
+		WaitForCancellation: true,
+	}
+	ctx = workflow.WithChildOptions(ctx, cwo)
+
+	var isProcessingDone = false
+
+	for !isProcessingDone {
+		ctx, cancelHandler := workflow.WithCancel(ctx)
+
+		// it is important to re-execute the child workflow after every signal
```
cretz, line 31: But you don't execute after every signal. If you get two signals you only execute after the latest. (seems to be by intention, just need to change this comment)

cancel-in-progress/parent_workflow.go
```diff
@@ -0,0 +1,109 @@
...
+			if err != nil {
+				logger.Error("Child execution failure.", "Error", err)
+			}
+
+			// Child workflow completed. Let's end the parent workflow.
+			isProcessingDone = true
+		})
+
+		selector.AddReceive(reBuildSignalChan, func(c workflow.ReceiveChannel, more bool) {
+			logger.Info("Received signal.", "Message", message)
```
cretz, line 55: This log shows the `message` from the _last_ time the loop ran. The message is not received until `GetLatestMessageFromChannel` is executed.

cancel-in-progress/parent_workflow.go
```diff
@@ -0,0 +1,109 @@
...
+			// drain the channel to get the latest signal
+			// Users might send multiple signals in a short period of time
+			// and we are only interested in the latest signal
+			message = GetLatestMessageFromChannel(logger, reBuildSignalChan)
+		})
+
+		// wait for the build workflow to finish or the signal to cancel the running execution through a signal
+		selector.Select(ctx)
+
+		// in case of an error we want to cancel the child workflow
```
cretz, line 70: By default, due to parent close policy, returning from the parent is actually _terminating_ the child, not cancelling it

cancel-in-progress/starter/main.go
```diff
@@ -0,0 +1,78 @@
...
+	"strconv"
+	"time"
+
+	cancel_in_progress "github.com/temporalio/samples-go/cancel-in-progress"
+)
+
+func main() {
+	// The client is a heavyweight object that should be created only once per process.
+	c, err := client.Dial(client.Options{
+		HostPort: client.DefaultHostPort,
```
cretz, line 17: The default can be left unset

StarpTech:
@cretz great review. I will address this soon.

PR #233, opened 2022-12-06
Add multi history replay sample

Quinn-With-Two-Ns:
Add multi history replay sample

### DIFF
--- a/multi-history-replay/replayer/main.go
+++ b/multi-history-replay/replayer/main.go
@@ -1,0 +1,78 @@
+package main
+
+import (
+	"context"
+	"log"
+
+	"github.com/temporalio/samples-go/helloworld"
+	"go.temporal.io/api/enums/v1"
+	"go.temporal.io/api/history/v1"
+	"go.temporal.io/api/workflow/v1"
+	"go.temporal.io/api/workflowservice/v1"
+	"go.temporal.io/sdk/client"
+	"go.temporal.io/sdk/worker"
+)
+
+func GetWorkflowHistory(ctx context.Context, client client.Client, id, runID string) (*history.History, error) {
+	var hist history.History
+	iter := client.GetWorkflowHistory(ctx, id, runID, false, enums.HISTORY_EVENT_FILTER_TYPE_ALL_EVENT)
+	for iter.HasNext() {
+		event, err := iter.Next()
+		if err != nil {
+			return nil, err
+		}
+		hist.Events = append(hist.Events, event)
+	}
+	return &hist, nil
+}
+
+// GetWorkflows calls ListWorkflow with query and gets as many workflow executions until there are no more or we the number exceeds maxWorkflows
+func GetWorkflows(ctx context.Context, c client.Client, query string, maxWorkflows int) ([]*workflow.WorkflowExecutionInfo, error) {
+	var nextPageToken []byte
+	workflowExecutions := make([]*workflow.WorkflowExecutionInfo, 0)
+	for {
+		resp, err := c.ListWorkflow(ctx, &workflowservice.ListWorkflowExecutionsRequest{
+			Query:         query,
+			NextPageToken: nextPageToken,
+		})
+		if err != nil {
+			return nil, err
+		}
+		workflowExecutions = append(workflowExecutions, resp.Executions...)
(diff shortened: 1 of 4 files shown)


Sushisource:
Thanks for doing this!

In the [Python sample](https://github.com/temporalio/samples-python/commit/8407dca0921cee3da825d94634c55c7ac2330e85), I separate the worker & replayer into separate files / processes which I feel is slightly more realistic in terms of how users actually use the replayer. 

I think that might worth be doing here too.

Quinn-With-Two-Ns:
OK split the `replayer` into a new file and made the repalyer use the `ListWorkflow` API

multi-history-replay/replayer/main.go
```diff
@@ -0,0 +1,78 @@
...
+	workflowExecutions, err := GetWorkflows(ctx, c, query, 10)
+	if err != nil {
+		log.Fatalln("Error listing workflows", err)
+	}
+	log.Println("Found workflows", "Count", len(workflowExecutions))
+
+	replayer := worker.NewWorkflowReplayer()
+	replayer.RegisterWorkflow(helloworld.Workflow)
+	for _, we := range workflowExecutions {
+		histroy, err := GetWorkflowHistory(ctx, c, we.Execution.GetWorkflowId(), we.Execution.GetRunId())
```
cretz, line 69:
```suggestion
		history, err := GetWorkflowHistory(ctx, c, we.Execution.GetWorkflowId(), we.Execution.GetRunId())
```

multi-history-replay/replayer/main.go
```diff
@@ -0,0 +1,78 @@
...
+		resp, err := c.ListWorkflow(ctx, &workflowservice.ListWorkflowExecutionsRequest{
+			Query:         query,
+			NextPageToken: nextPageToken,
+		})
+		if err != nil {
+			return nil, err
+		}
+		workflowExecutions = append(workflowExecutions, resp.Executions...)
+		nextPageToken = resp.NextPageToken
+		if nextPageToken == nil || len(workflowExecutions) >= maxWorkflows {
```
cretz, line 43:
```suggestion
		if len(nextPageToken) == 0 || len(workflowExecutions) >= maxWorkflows {
```

Would rather not trust proto internals get nil vs empty right (and usually len-zero is a better way to check for nil byte slice in Go unless there's a real reason to differentiate between that and empty)

multi-history-replay/replayer/main.go
```diff
@@ -0,0 +1,78 @@
...
+		}
+		hist.Events = append(hist.Events, event)
+	}
+	return &hist, nil
+}
+
+// GetWorkflows calls ListWorkflow with query and gets as many workflow executions until there are no more or we the number exceeds maxWorkflows
+func GetWorkflows(ctx context.Context, c client.Client, query string, maxWorkflows int) ([]*workflow.WorkflowExecutionInfo, error) {
+	var nextPageToken []byte
+	workflowExecutions := make([]*workflow.WorkflowExecutionInfo, 0)
```
cretz, line 32:
```suggestion
	var workflowExecutions []*workflow.WorkflowExecutionInfo
```

Doesn't really matter I just happen to like this one better. But it is subjective (perf diff negligible), so your choice.

cretz (approved):
Looks great! Would like @Sushisource to approve before merging to ensure it has what he was expecting.

multi-history-replay/replayer/main.go
```diff
@@ -0,0 +1,61 @@
...
+	"github.com/temporalio/samples-go/helloworld"
+	workflowpb "go.temporal.io/api/workflow/v1"
+	"go.temporal.io/api/workflowservice/v1"
+	"go.temporal.io/sdk/client"
+	"go.temporal.io/sdk/worker"
+	"go.temporal.io/sdk/workflow"
+)
+
+// GetWorkflows calls ListWorkflow with query and gets all workflow exection infos in a list.
+func GetWorkflows(ctx context.Context, c client.Client, query string) ([]*workflowpb.WorkflowExecutionInfo, error) {
```
Sushisource, line 16: Haha oops we added list workflows :smile:

Sushisource (approved):
Looking good. Thanks Quinn!

PR #235, opened 2022-12-09
Split merge

mfateev:
## What was changed
Replaced `parallel` and `splitmerge` samples as they confused users. In most cases, use of `workflow.Go` to parallelize activity and child workflow executions is not the simplest approach.

I've added three samples:
* `goroutine` that demonstrates the use of `workflow.Go` to parallelize calling a sequence of operations.
* `splitmerge-future` to demonstrate how to implement splitmerge using Future.Get.
* `splitmerge-selector` to demonstrate how to implement splitmerge using Selector to process results as they become available.


goroutine/goroutine_workflow.go
```diff
@@ -0,0 +1,62 @@
...
+	ao := workflow.ActivityOptions{
+		StartToCloseTimeout: 10 * time.Second,
+	}
+	ctx = workflow.WithActivityOptions(ctx, ao)
+
+	for i := 0; i < parallelism; i++ {
+		input1 := fmt.Sprint(i) // Should be outside lambda to be captured correctly
+		// Start a goroutine in a workflow safe way
+		workflow.Go(ctx, func(gCtx workflow.Context) {
+			// It is important to use the context passed to the goroutine function
```
cretz, line 28: > It is important to use the context passed to the goroutine function

I usually shadow the outer variable by giving the parameter the same name (`ctx` in this case) to avoid mistakes. But unsure if doing so detracts from the lesson.

goroutine/goroutine_workflow.go
```diff
@@ -0,0 +1,62 @@
...
+		})
+	}
+
+	// Wait for Goroutines to complete. Await blocks until the condition function returns true.
+	// The function is evaluated on every workflow state change. Consider using `workflow.AwaitWithTimeout` to
+	// limit duration of the wait.
+	workflow.Await(ctx, func() bool {
+		return err != nil || len(results) == parallelism
+	})
+	return results, err
```
cretz, line 51:
```suggestion
	return
```

(granted this is a bit subjective)

cretz (approved):
LGTM. Will let @Quinn-With-Two-Ns have a bit to review too if he'd like since he's also donig a lot of Go work these days.

Quinn-With-Two-Ns (approved):

goroutine/goroutine_workflow.go
```diff
@@ -0,0 +1,62 @@
...
+				return
+			}
+			results = append(results, result2)
+		})
+	}
+
+	// Wait for Goroutines to complete. Await blocks until the condition function returns true.
+	// The function is evaluated on every workflow state change. Consider using `workflow.AwaitWithTimeout` to
+	// limit duration of the wait.
+	workflow.Await(ctx, func() bool {
```
cretz, line 48: CI linter caught a missing error result check here

PR #236, opened 2022-12-09
Added await-signals sample

mfateev:
## What was changed
Added a sample demonstrating how to process signals that come out of order using `Await` and `AwaitWithTimeout`. 

## Why?
The presented pattern greatly simplifies workflow implementation in many scenarios.

### DIFF
--- a/await-signals/await_signals_workflow.go
+++ b/await-signals/await_signals_workflow.go
@@ -1,0 +1,164 @@
+package await_signals
+
+import (
+	"go.temporal.io/sdk/temporal"
+	"go.temporal.io/server/common"
+	"time"
+
+	"go.temporal.io/sdk/workflow"
+)
+
+/**
+ * The sample demonstrates how to deal with multiple signals that can come out of order and require actions
+ * if a certain signal not received in a specified time interval.
+ *
+ * This specific sample receives three signals: Signal1, Signal2, Signal3. They have to be processed in the
+ * sequential order, but they can be received out of order.
+ * There are two timeouts to enforce.
+ * The first one is the maximum time between signals.
+ * The second limits the total time since the first signal received.
+ *
+ * A naive implementation of such use case would use a single loop that contains a Selector to listen on three
+ * signals and a timer. Something like:
+
+ *	for {
+ *		selector := workflow.NewSelector(ctx)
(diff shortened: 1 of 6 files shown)


await-signals/await_signals_workflow.go
```diff
@@ -0,0 +1,164 @@
+package await_signals
+
+import (
+	"go.temporal.io/sdk/temporal"
+	"go.temporal.io/server/common"
```
cretz, line 5: We shouldn't have a dependency on the server repo in this repo at all IMO. We only did for samples that require access to the CLI plugin interface (happened before CLI moved out of server) and in a couple of cases it was used for its utilities (unnecessarily IMO).

I opened https://github.com/temporalio/samples-go/issues/237 for this. Don't have to fix now.

await-signals/await_signals_workflow.go
```diff
@@ -0,0 +1,164 @@
...
+	}
+	log.Info("Signal1 Processed")
+
+	// Wait for Signal2
+	timeout, err := a.GetNextTimeout(ctx)
+	// No time left. At this point this cannot really happen.
+	if err != nil {
+		return
+	}
+	ok, err := workflow.AwaitWithTimeout(ctx, timeout, func() bool {
```
cretz, line 132: Also a good use case for your new `ReceiveWithTimeout`, but I understand that doesn't demonsrtate what you want here

await-signals/await_signals_workflow.go
```diff
@@ -0,0 +1,164 @@
...
+	return common.MinDuration(SignalToSignalTimeout, totalLeft), nil
+}
+
+// AwaitSignalsWorkflow workflow definition
+func AwaitSignalsWorkflow(ctx workflow.Context) (err error) {
+	log := workflow.GetLogger(ctx)
+	var a AwaitSignals
+	workflow.Go(ctx, func(ctx workflow.Context) {
+		a.Listen(ctx)
+	})
```
cretz, line 114:
```suggestion
	workflow.Go(ctx, a.Listen)
```

cretz (approved):

await-signals/await_signals_workflow.go
mfateev: oops, missed that it is not built-in.

PR #240, opened 2022-12-28
RequestCancelExternalWorkflow Sample

mnichols:
I was trying to figure out the best way to test cancellations of Child Workflows. In the past, only cancelling the context was possible so I thought I'd try out `RequestCancelExternalWorkflow`.

I stumbled on some seeming inconsistencies I thought I'd highlight to verify whether a bug exists or not.

There are two tests:

- `Test_Cancel_NoMocky` which does not mock out ChildWorkflow execution. The `SetOnChildWorkflowCanceledListener` _is_ invoked. It is noteworthy that using `OnRequestCancelExternalWorkflow` fails, which I suppose makes sense but we should make clearer in docs that this API is only for mocked ChildWorkflow calls. 

- `Test_Cancel_WithMocky` mocks the `MyChild` workflow. The `SetOnChildWorkflowCanceledListener` _is not_ invoked, though the `SetOnChildWorkflowCompletedListener` _is_ invoked. This seems inconsistent to me. I ack that when someone uses `OnRequestCancelExternalWorkflow` they should Assert the expected call on that...however since tests evolve in projects and mocking is eventually  introduced for children so the cancel listener no longer being called is hard to understand. We should at least call out that the listener is not called when using this API.

This sample shows how to use the `RequestCancelExternalWorkflow` API. 
There are some nuances using this not very obvious (to me):
1. I observed that workflow tasks would fail with `Unhandled Command` errors (appear in workflow history) when the ChildWorkflow future was not resolved. Doing a simple `.Get` after calling `RequestCancelExternalWorkflowExecution` fixed this.
2. Testing when using this API was not as straightforward as I'd hoped (see above).
3. I found a bug when using `OnWorkflow` mock for workflows that do not have arguments...you _must_ still pass in an argument (eg `mock.Anything`) or else the `workflowInterceptor` panics...seems like we are missing a check for params before spreading them somewhere.

### DIFF
--- a/request-cancel-external-workflow/workflow_test.go
+++ b/request-cancel-external-workflow/workflow_test.go
@@ -1,0 +1,183 @@
+package requestcancelexternalworkflow
+
+import (
+	"github.com/stretchr/testify/mock"
+	"github.com/stretchr/testify/suite"
+	"go.temporal.io/sdk/converter"
+	"go.temporal.io/sdk/log"
+	"go.temporal.io/sdk/testsuite"
+	"go.temporal.io/sdk/workflow"
+	"testing"
+	"time"
+)
+
+const childWorkflowID string = "luke"
+const signalName string = "mysignal"
+const signalValueDone string = "done"
+
+func MyWorkflow(ctx workflow.Context) error {
+	ao := workflow.ActivityOptions{
+		StartToCloseTimeout: 30 * time.Minute,
+		HeartbeatTimeout:    5 * time.Second,
+		WaitForCancellation: true,
+	}
+	ctx = workflow.WithActivityOptions(ctx, ao)
+	info := workflow.GetInfo(ctx)
+
+	logger := log.With(workflow.GetLogger(ctx), "type", info.WorkflowType, "wid", info.WorkflowExecution.ID)
+	logger.Info("workflow started", "info", workflow.GetInfo(ctx))
+
+	childCtx := workflow.WithChildOptions(ctx, workflow.ChildWorkflowOptions{
+		WorkflowID:          childWorkflowID,
+		WaitForCancellation: true,
+	})
+
+	var signalValue string
+	var childFuture workflow.Future
+	childFuture = workflow.ExecuteChildWorkflow(childCtx, MyChild)
+	sigChan := workflow.GetSignalChannel(ctx, signalName)
+	sel := workflow.NewSelector(ctx)
+	workflow.Go(ctx, func(ctx workflow.Context) {
+		sel.AddReceive(sigChan, func(c workflow.ReceiveChannel, more bool) {
+			c.Receive(ctx, &signalValue)
+			logger.Info("signal received", "signal", signalName, signalValue)
+		})
+		sel.Select(ctx)
+	})
+
+	if err := workflow.Await(ctx, func() bool {
+		return signalValue == signalValueDone
+	}); err != nil {
+		logger.Error("await err", "err", err)
+	}
+	logger.Info("released from block")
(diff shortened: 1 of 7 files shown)


request-cancel-external-workflow/workflow_test.go
```diff
@@ -0,0 +1,183 @@
...
+}
+func (s *CancelTestSuite) Test_Cancel_WithMocky() {
+
+	s.env.RegisterWorkflow(MyWorkflow)
+	s.env.RegisterWorkflow(MyChild)
+	//var cancelRequestCalled bool
+	var childCancelInfo *workflow.Info
+	var childCompletedInfo *workflow.Info
+	// this mock is broken...MyChild does not require any arguments but the workflowInterceptor in our test suite panics without args
+	s.env.OnWorkflow(MyChild, mock.Anything).Return(func(ctx workflow.Context) error {
```
mnichols, line 159: The workflow definition does not have arguments, but our mock seems to panic when you don't provide any, so using `mock.Anything` made the problem go away. (bug?)

request-cancel-external-workflow/workflow_test.go
```diff
@@ -0,0 +1,183 @@
...
+		return nil
+	})
+	s.env.SetOnChildWorkflowCanceledListener(func(workflowInfo *workflow.Info) {
+		childCancelInfo = workflowInfo
```
mnichols, line 163: This does not get called, presumably because we are mocking the ChildWorkflow...but note that the `SetOnChildWorkflowCompletedListener` _does_ get called, so it is a little surprising (not obvious).

request-cancel-external-workflow/workflow_test.go
```diff
@@ -0,0 +1,183 @@
...
+	})
+	s.env.SetOnChildWorkflowCompletedListener(func(workflowInfo *workflow.Info, result converter.EncodedValue, err error) {
+		childCompletedInfo = workflowInfo
+	})
+	s.env.OnRequestCancelExternalWorkflow(mock.Anything, childWorkflowID, "").Return(nil).Once()
```
mnichols, line 168: You must assert on this if you are mocking a child workflow and want to verify the cancellation of it. The cancellation listener does not get called.

request-cancel-external-workflow/workflow_test.go
```diff
@@ -0,0 +1,183 @@
...
+
+	s.env.RegisterDelayedCallback(func() {
+		s.env.SignalWorkflowByID("default-test-workflow-id", signalName, signalValueDone)
+	}, time.Second*1)
+	s.env.ExecuteWorkflow(MyWorkflow)
+	s.True(s.env.IsWorkflowCompleted())
+	s.NoError(s.env.GetWorkflowError())
+	//s.True(cancelRequestCalled)
+	s.NotNil(childCancelInfo)
```
mnichols, line 177: This fails because the cancellation listener isnt invoked.

request-cancel-external-workflow/workflow_test.go
Quinn-With-Two-Ns: I think the first mock argument here is for the context so not a bug

mnichols: ah good point..i forgot about that

request-cancel-external-workflow/workflow_test.go
Quinn-With-Two-Ns: Looking at the go test suite code I think the current behavior is correct. Right now the workflow assumes the child WF will be blocked until it is signaled so when you call `RequestCancelExternalWorkflow` you know it will cancel the child WF. In this test since you mock the child WF, the child WF will instantly complete so your call to `RequestCancelExternalWorkflow` will not cancel the child WF since it is already complete (not running) so you don't get the callback.

PR #242, opened 2023-01-03
move to golangci-lint and address linting issues

miku:
## What was changed

Previously, two separate checks "errcheck" and "staticcheck" were used. These are also part of "golangci-lint" [1] metalinter.

In addition, "golangci-lint" has a nolint directive [2], which at least "errcheck" did not seem to have.

At this point in time, "golangci-lint" bundles 72 linters [3] (of which this project uses two) - so additional check can easily be added. A configuration file was required to suppress false positives regarding determinism stemming from "print" and "fmt" functions.

As this project use "go 1.16" some safe linting suggestions have been applied (e.g. related to deprecation of https://pkg.go.dev/io/ioutil).

* [1] https://golangci-lint.run/
* [2] https://golangci-lint.run/usage/false-positives/#nolint-directive
* [3] https://golangci-lint.run/usage/linters/

## Why?

I wanted to use the [nolint directive](https://golangci-lint.run/usage/false-positives/#nolint-directive) of [golangci-lint](https://golangci-lint.run/) to selectively ignore checks, e.g. to ignore an error in `defer` when closing an HTTP response body.  

At the same time, it seemed to simplify the `Makefile` to use a single "metalinter" instead of two separate linters.

## Checklist

2. How was this tested:

Tested locally, via:

```
$ make
```

3. Any docs updates needed?

Not required.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
 ############################# Main targets #############################
 # Run all checks, build, and test.
-install: clean staticcheck errcheck workflowcheck bins test
+install: clean lint workflowcheck bins test
 ########################################################################
 
 ##### Variables ######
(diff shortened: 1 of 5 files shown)


expense/activities.go
```diff
@@ -57,8 +57,8 @@ func WaitForDecisionActivity(ctx context.Context, expenseID string) (string, err
 		logger.Info("waitForDecisionActivity failed to register callback.", "Error", err)
 		return "", err
 	}
-	body, err := ioutil.ReadAll(resp.Body)
-	_ = resp.Body.Close()
+	defer resp.Body.Close() // nolint:errcheck
+	body, err := io.ReadAll(resp.Body)
```
cretz, line 61: I think we should leave this how it was and just change `ioutil` to `os`. We want to explicitly show we ignore errors and there is no benefit to changing this to a defer here.

Makefile
```diff
@@ -27,20 +27,15 @@ test:
...
+lint:
+	@printf $(COLOR) "Run checks..."
+	@go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.50.1
+	@golangci-lint run --disable-all -E errcheck -E staticcheck
 
 workflowcheck:
 	@printf $(COLOR) "Run workflow check..."
 	@go install go.temporal.io/sdk/contrib/tools/workflowcheck
-	@workflowcheck -show-pos ./...
+	@workflowcheck -config workflowcheck.config.yaml -show-pos ./...
```
cretz, line 38: Why add this config? Is something triggering it?

cretz (commented):
Thanks! I think we should do the same to https://github.com/temporalio/sdk-go then too. It's a bit off to do it one way in our samples but another in the canonical Go SDK project. (some in-progress PRs are already changing from using ioutil, so we may have to wait until after those)

PR #243, opened 2023-01-03
adding basic example to work with dev guide foundations

flossypurse:
## What was changed

Added a new sample `/yourapp`

## Why?

This sample is meant to provide snipped examples for foundations section of the Dev guide

**PREVIEW:** https://github.com/temporalio/documentation/pull/1868

It combines points made in both hello world and the basic greeting apps, but rename functions and parameters to match what we use in the guide.

## Checklist

1. Closes 

n/a

2. How was this tested:

There is a basic replay test and basic workflow and activity tests.

TODO to make them better and usable(snippable) for the guide.

3. Any docs updates needed?

Will be used as snipped examples in the dev guide foundations page for Go: https://docs.temporal.io/application-development/foundations?lang=go


yourapp/your_workflow_definition.go
```diff
@@ -0,0 +1,60 @@
...
+	// Not specifying a Task Queue will default to the parent Workflow Task Queue.
+	activityOptions := workflow.ActivityOptions{
+		StartToCloseTimeout: 10 * time.Second,
+	}
+	ctx = workflow.WithActivityOptions(ctx, activityOptions)
+	activityParam := YourActivityParam{
+		ActivityParamX: param.WorkflowParamX,
+		ActivityParamY: param.WorkflowParamY,
+	}
+	// Use a nil struct pointer to call Activities that are part of a struct
```
cretz, line 33: Inconsistent with trailing punctuation on comments

yourapp/your_activity_definition.go
```diff
@@ -0,0 +1,50 @@
...
+// If the Worker crashes this Activity object loses its state.
+type YourActivityObject struct {
+	SharedMessageState string
+	SharedCounterState int
+}
+// An Activity Definiton is an exportable function.
+func(a *YourActivityObject) YourActivityDefinition(ctx context.Context, param YourActivityParam) (YourActivityResultObject, error) {
+	// Use Acivities for computations or calling external APIs.
+	// This is just an example of appending to text and incrementing a counter.
+	a.SharedMessageState = param.ActivityParamX + " World!"
```
cretz, line 29: This is usually bad as this is not concurrency safe and shared state from an activity should be

yourapp/your_activity_definition.go
```diff
@@ -0,0 +1,50 @@
...
+}
+
+func(a *YourActivityObject) PrintSharedSate(ctx context.Context) error {
+	logger := activity.GetLogger(ctx)
+	logger.Info("The current message is:", a.SharedMessageState)
+	logger.Info("The current counter is:", a.SharedCounterState)
+	return nil
+}
+// An Activity Definiton is an exportable function.
+func YourSimpleActivityDefinition(ctx context.Context) error {
```
cretz, line 47: Is this used anywhere?

yourapp/your_workflow_definition.go
```diff
@@ -0,0 +1,60 @@
...
+	}
+	// Make the results of the Workflow Execution available.
+	workflowResult := YourWorkflowResultObject {
+		ResultFieldX: activityResult.ResultFieldX,
+		ResultFieldY: activityResult.ResultFieldY,
+	}
+	return workflowResult, nil
+}
+
+func YourSimpleWorkflowDefinition(ctx workflow.Context) error {
```
cretz, line 57: Is this used anywhere?

yourapp/your_workflow_definition_replay_test.go
```diff
@@ -0,0 +1,44 @@
...
+	"github.com/stretchr/testify/require"
+	"github.com/stretchr/testify/suite"
+	"go.temporal.io/api/workflowservicemock/v1"
+	"go.temporal.io/sdk/worker"
+)
+
+type replayTestSuite struct {
+	suite.Suite
+	mockCtrl *gomock.Controller
+	service  *workflowservicemock.MockWorkflowServiceClient
```
cretz, line 16: Why have this mock stuff? Actually, why have a test suite at all instead of just simple test function?

yourapp/your_workflow_definition_replay_test.go
```diff
@@ -0,0 +1,44 @@
...
+// This replay test is the recommended way to make sure changing workflow code is backward compatible without non-deterministic errors.
+// "yourworkflow.json" can be downloaded from Temporal CLI:
+//      tctl wf show -w your-workflow-id --output_filename ./your-workflow-id.json
+// Or from Temporal Web UI. And you may need to change workflowType in the first event.
+func (s *replayTestSuite) TestReplayWorkflowHistoryFromFile() {
+	replayer := worker.NewWorkflowReplayer()
+
+	replayer.RegisterWorkflow(YourWorkflowDefinition)
+
+	err := replayer.ReplayWorkflowHistoryFromJSONFile(nil, "yourworkflow.json")
```
cretz, line 42: If tests are run from another directory, this may fail. You can `go:embed` or you can do something like `_, currFile, _, _ := runtime.Caller(0)` + `filepath.Dir(currFile)`

yourapp/README.md
```diff
@@ -0,0 +1,10 @@
+### Steps to run this sample:
+1) You need a Temporal service running. See details in README.md
+2) Run the following command to start the worker
+```
+go run yourworkflow/worker/main.go
+```
+3) Run the following command to start an HTTP server
```
cretz, line 7: What do I do after starting an HTTP server?

yourapp/your_activity_definition.go
flossypurse: Shared state field, should it be a pointer then?

yourapp/your_activity_definition.go
flossypurse: No, I was considering snipping it though to show it is valid compiled code.
And poissibly registering it with the Worker...

yourapp/your_workflow_definition.go
flossypurse: Same with this, not yet, I was considering snipping it to show it is valid compiled code and possibly registering it with the Worker.

yourapp/your_workflow_definition_replay_test.go
flossypurse: tbh, I just copied the tests that were in the [helloworld example](https://github.com/temporalio/samples-go/blob/main/helloworld/replay_test.go) and made them work.

What would you recommend?

yourapp/gateway/main.go
```diff
@@ -0,0 +1,55 @@
...
+	})
+	if err != nil {
+		log.Fatalln("Unable to create Temporal Client", err)
+	}
+	defer temporalClient.Close()
+	// Start an HTTP server and listen on /start
+	http.HandleFunc("/start", func(w http.ResponseWriter, r *http.Request) {
+		startWorkflowHandler(w, r, temporalClient)
+	})
+	http.ListenAndServe(":8081", nil)
```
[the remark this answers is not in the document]
flossypurse: We actually hear a lot from newer users that seeing the temporal client used like this immediately helps to orient themselves around the model of the system.

Seems like pretty low overhead to accommodate wdyt?

And users can start WFs with curl commands rather than running a main.go file (yes it is 1 more step) 
But still, the value add for newer users could be quite big I hear.

yourapp/your_activity_definition.go
```diff
@@ -0,0 +1,50 @@
...
+	ResultFieldX string
+	ResultFieldY int
+}
+// If the Worker crashes this Activity object loses its state.
+type YourActivityObject struct {
+	SharedMessageState string
+	SharedCounterState int
+}
+// An Activity Definiton is an exportable function.
+func(a *YourActivityObject) YourActivityDefinition(ctx context.Context, param YourActivityParam) (YourActivityResultObject, error) {
```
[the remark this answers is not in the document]
flossypurse: yes, did that now.

yourapp/your_workflow_definition_replay_test.go
flossypurse: I tested your statement and ran the tests from a bunch of other directories without issue - so... this seems to be just fine?

yourapp/your_workflow_definition_replay_test.go
flossypurse: Reduced to simple test function.

yourapp/your_activity_definition.go
cretz: No, you need to surround it with locks.

Are we expecting to have samples in other languages just like this one? I am curious what the end result may be.

yourapp/your_activity_definition.go
cretz: But this is sample code. We don't want dead code in our samples to support another use case. Maybe code doesn't have to be in samples for documentation use.

yourapp/your_workflow_definition.go
cretz: Mentioned above, yeah this may not be good sample material. Our samples have use cases about proper Go code and dead/unused code is bad.

yourapp/your_workflow_definition_replay_test.go
cretz: Yes, it seems it sets the local directory. But usually users shouldn't count on local directory on their runs. Yes this is fine, but if users copy this code for use outside of tests (which many do for replayers) it'll break.

yourapp/README.md
```diff
@@ -0,0 +1,18 @@
+### Steps to run this sample:
```
cretz, line 1: What is the general purpose of this sample over the hello world sample? If it's just for docs purpose and not for user sample purpose, I wonder if it should live elsewhere. Or can you adapt the hello world sample?

Also, we name all of our sample folders according to what the sample demonstrates. We should do the same here.

yourapp/gateway/main.go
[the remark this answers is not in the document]
napcs: Most Go developers can also write a CLI, but we include that because that's not the point - the point is showing how to start a workflow execution from code. What we need to show is specifically and explicitly HOW a developer uses Temporal in the context of a different kind of application. This comes up so often. I can assure you it's not as obvious to people new to Temporal.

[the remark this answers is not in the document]
cretz: > What we need to show is specifically and explicitly HOW a developer uses Temporal in the context of a different kind of application.

Then make that the focal point. No need to litter _basic_ samples with HTTP servers. How am I supposed to know that the `/yourapp` sample shows me how to call Temporal from an HTTP server?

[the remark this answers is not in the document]
napcs: > Then make that the focal point. No need to litter basic samples with HTTP servers

Agree here. We should have *an* example of this pattern.

Issue #239, opened 2022-12-21
[Feature Request] Porting polling samples from Java examples


edmondop:
### Is your feature request related to a problem? Please describe.

Java examples of SDK usage contain an useful example of how to implement polling activities:

- [Frequently Polling Activity](https://github.com/temporalio/samples-java/blob/main/src/main/java/io/temporal/samples/polling/frequent/README.md)
- [Infrequently Polling Activity](https://github.com/temporalio/samples-java/blob/main/src/main/java/io/temporal/samples/polling/infrequent/README.md)
- [Periodic Polling of a sequence of activities](https://github.com/temporalio/samples-java/blob/main/src/main/java/io/temporal/samples/polling/periodicsequence/README.md)

The samples are based on [this](https://community.temporal.io/t/what-is-the-best-practice-for-a-polling-activity/328/2) 
community forum thread.

### Describe the solution you'd like

It would be very useful to bring them to Go examples as well.

tsurdilo:
Thanks @edmondop !

edmondop:
@tsurdilo after reviewing the code in the Java repository (https://github.com/temporalio/samples-java/tree/main/src/main/java/io/temporal/samples/polling), it seems there are three different polling-based scenarios:

- frequent polling activity
- infrequent polling activity
- periodic polling of a sequence of activities

They are organized within a single java package. GO examples have a different organization, i.e. each folder is a single scenario/example. How do you think I should proceed?

- Option A: having a parent folder "polling" containing three subfolders (frequent,infrequent, sequence)
- Option B: having three folders polling_infrequent, polling_frequent, polling_sequence

tsurdilo:
@edmondop i think if it is possible to do Option A that would be better. if its not possible Option B is fine too.

edmondop:
@tsurdilo one additional question, I was looking to https://github.com/temporalio/samples-java/blob/ae302763df044d2d3c62039ac8fc2658fd418daa/src/main/java/io/temporal/samples/polling/periodicsequence/PeriodicPollingStarter.java and I am really struggling to understand how the parent workflow is invoked, given it is not registered.

Nor in the test that workflow is invoked explicitly, only the parent workflow is invoked. I found that kind of confusing, does the parent workflow has a role or is it dead code?

edmondop:
Fixed with #246

PR #246, opened 2023-01-16
Porting to GoLang java Polling Examples

edmondop:
## What was changed
Addresses #239 by creating a three examples within the polling module

## Why?

Because I want to learn GoLang

## Checklist

- [ ] clarify how to use code snippet for documentation


polling/frequent/activities.go
```diff
@@ -0,0 +1,32 @@
...
+// DoPoll Activity.
+// In this activity polling is implemented within the activity itself and not the workflow,
+// using the heartbeat mechanism to keep the activity alive
+func (a *PollingActivities) DoPoll(ctx context.Context) (string, error) {
+	for {
+		res, err := a.TestService.GetServiceResult()
+		if err == nil {
+			return res, err
+		}
+		activity.RecordHeartbeat(ctx, "no response")
```
mfateev, line 26: I would not provide details in this case as they are intended to record the activity progress. In the future, we might skip a DB write if the details are empty.

polling/frequent/workflow.go
```diff
@@ -0,0 +1,29 @@
...
+	"time"
+
+	"go.temporal.io/sdk/workflow"
+)
+
+func FrequentPolling(ctx workflow.Context) (string, error) {
+	logger := workflow.GetLogger(ctx)
+
+	ao := workflow.ActivityOptions{
+		StartToCloseTimeout: 10 * time.Second,
```
mfateev, line 13: StartToClose timeout should be set to a very long value as the activity is expected to be as long-running as needed for the resource to become available. 
HeartbeatTimeout must be specified for activity heartbeat to be ever recorded. This also allows restarting an activity quickly after its worker failure.

polling/periodic_sequence/child_workflow.go
```diff
@@ -0,0 +1,39 @@
...
+
+	"go.temporal.io/sdk/workflow"
+)
+
+type ChildWorkflowParams struct {
+	SingleWorkflowAttempts int
+	PollingInterval        time.Duration
+}
+
+func ChildWorkflow(ctx workflow.Context, params ChildWorkflowParams) (string, error) {
```
mfateev, line 14: I would rename it something like "PollingChildWorkflow". Just Child doesn't convey its purpose.

polling/periodic_sequence/child_workflow.go
```diff
@@ -0,0 +1,39 @@
+package periodic_sequence
+
+import (
+	"time"
+
+	"go.temporal.io/sdk/workflow"
+)
+
+type ChildWorkflowParams struct {
+	SingleWorkflowAttempts int
```
mfateev, line 10: I think SingleWorkflowAttempts is an implementation detail of the child. So there is no reason for the parent to set it.

polling/periodic_sequence/child_workflow.go
edmondop: 👍

polling/frequent/activities.go
edmondop: Makes sense, should I also update the Java example here? https://github.com/temporalio/samples-java/blob/main/src/main/java/io/temporal/samples/polling/frequent/FrequentPollingActivityImpl.java#L51

polling/frequent/workflow.go
edmondop: I have adopted the same StartToCloseTimeout of the java example, do you think it should be higher in Java too? https://github.com/temporalio/samples-java/blob/main/src/main/java/io/temporal/samples/polling/frequent/FrequentPollingWorkflowImpl.java#L40

polling/periodic_sequence/child_workflow.go
edmondop: 👍 I thought it could be useful for unit testing purposes to have control around it, but in reality controlling the polling interval is more important to have unit tests that execute quickly.  This was before I added the PollingDuration to the struct. 

Do you think hardcoding in the child workflow is appropriate? Making it private and default it via constructor-like factory function would not work, I believe, because it won't be serialized

polling/frequent/workflow.go
mfateev: Yes, the Java sample should be fixed as well.

edmondop: I have opened https://github.com/temporalio/samples-java/pull/411 for the issue in the Java samples

polling/frequent/activities.go
edmondop: https://github.com/temporalio/samples-java/pull/412

polling/frequent/workflow.go
```diff
@@ -0,0 +1,30 @@
...
+import (
+	"time"
+
+	"go.temporal.io/sdk/workflow"
+)
+
+func FrequentPolling(ctx workflow.Context) (string, error) {
+	logger := workflow.GetLogger(ctx)
+
+	ao := workflow.ActivityOptions{
```
mfateev, line 12: These timeouts are too short. I would change them to StartToClose=24 hours and Heartbeat=30 seconds. We need to update the Java sample as well.

polling/frequent/workflow.go
```diff
@@ -0,0 +1,30 @@
...
+		StartToCloseTimeout: 60 * time.Second,
+		HeartbeatTimeout:    2 * time.Second,
+	}
+	ctx = workflow.WithActivityOptions(ctx, ao)
+
+	// @@@SNIPSTART samples-go-polling-frequent-activity
```
mfateev, line 18: I don't think we need SNIP comments as these are not referenced from documentation.

polling/frequent/workflow.go
```diff
@@ -0,0 +1,30 @@
...
+	var a *PollingActivities // use a nil struct pointer to call activities that are part of a structure
+
+	var pollResult string
+	err := workflow.ExecuteActivity(ctx, a.DoPoll).Get(ctx, &pollResult)
+	if err != nil {
+		logger.Error("Get greeting failed.", "Error", err)
```
mfateev, line 24: Fix the name of the activity in the error message.

polling/infrequent/worker/main.go
```diff
@@ -0,0 +1,36 @@
...
+	// The client and worker are heavyweight objects that should be created once per process.
+	c, err := client.Dial(client.Options{
+		HostPort: client.DefaultHostPort,
+	})
+	if err != nil {
+		log.Fatalln("Unable to create client", err)
+	}
+	defer c.Close()
+
+	w := worker.New(c, "pollingSampleQueue", worker.Options{})
```
mfateev, line 23: Use a different task queue name for each sample. Otherwise, it might create confusion if a worker for another sample is still running.

polling/infrequent/workflow_test.go
```diff
@@ -0,0 +1,35 @@
...
+	testsuite.WorkflowTestSuite
+}
+
+func TestUnitTestSuite(t *testing.T) {
+	suite.Run(t, new(UnitTestSuite))
+}
+
+func (s *UnitTestSuite) Test_InfrequentPollingWorkflow() {
+	env := s.NewTestWorkflowEnvironment()
+	testService := polling.NewTestService(1) // Horrible workaround to avoid a timeout
```
mfateev, line 21: I don't think this is needed. I believe the test environment time skipping should allow running this test quickly even with many activity retries.

polling/periodic_sequence/child_workflow.go
```diff
@@ -0,0 +1,39 @@
...
+type ChildWorkflowParams struct {
+	PollingInterval time.Duration
+}
+
+func PollingChildWorkflow(ctx workflow.Context, params ChildWorkflowParams) (string, error) {
+	logger := workflow.GetLogger(ctx)
+	logger.Info("Starting child workflow with params", params)
+	ao := workflow.ActivityOptions{
+		StartToCloseTimeout:    10 * time.Second,
+		ScheduleToCloseTimeout: 2 * time.Second,
```
mfateev, line 18: ScheduleToClose timeout should never be smaller than StartToClose timeout as it is a timeout for overall activity execution including retries. I would remove ScheduleToClose timeout here. If you want to disable retries set the activity retry options.maxAttempts to 1.

polling/periodic_sequence/starter/main.go
```diff
@@ -0,0 +1,34 @@
...
+		HostPort: client.DefaultHostPort,
+	})
+	if err != nil {
+		log.Fatalln("Unable to create client", err)
+	}
+	defer c.Close()
+
+	workflowOptions := client.StartWorkflowOptions{
+		ID:        "pollingSampleQueue_" + uuid.New(),
+		TaskQueue: "pollingSampleQueue",
```
mfateev, line 26: Make sample specific.

mfateev (changes requested):

polling/frequent/activities.go
```diff
@@ -0,0 +1,32 @@
...
+func (a *PollingActivities) DoPoll(ctx context.Context) (string, error) {
+	for {
+		res, err := a.TestService.GetServiceResult()
+		if err == nil {
+			return res, err
+		}
+		activity.RecordHeartbeat(ctx)
+		time.Sleep(a.PollInterval)
+	}
+	return a.TestService.GetServiceResult()
```
cretz, line 29: Unreachable code

polling/frequent/activities.go
```diff
@@ -0,0 +1,32 @@
...
+// In this activity polling is implemented within the activity itself and not the workflow,
+// using the heartbeat mechanism to keep the activity alive
+func (a *PollingActivities) DoPoll(ctx context.Context) (string, error) {
+	for {
+		res, err := a.TestService.GetServiceResult()
+		if err == nil {
+			return res, err
+		}
+		activity.RecordHeartbeat(ctx)
+		time.Sleep(a.PollInterval)
```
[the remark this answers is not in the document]
cretz: :+1: Also send the context to `GetServiceResult()` to feel more Go like

polling/frequent/starter/main.go
```diff
@@ -0,0 +1,32 @@
...
+	"log"
+
+	"github.com/pborman/uuid"
+	"go.temporal.io/sdk/client"
+)
+
+func main() {
+	// The client is a heavyweight object that should be created once per process.
+	c, err := client.Dial(client.Options{
+		HostPort: client.DefaultHostPort,
```
cretz, line 15:
```suggestion
```

Redundant

polling/infrequent/workflow_test.go
cretz: There might be a default max attempts (10 or so?) to prevent spinning, but this could be set to something like 5.

polling/periodic_sequence/child_workflow.go
```diff
@@ -0,0 +1,39 @@
...
+	for i := 0; i < 10; i++ {
+		// Here we would invoke a sequence of activities
+		// For sample we just use a single one repeated several times
+		var pollResult string
+		err := workflow.ExecuteActivity(ctx, a.DoPoll).Get(ctx, &pollResult)
+		if err == nil {
+			return pollResult, nil
+		} else {
+			logger.Error("Error in activity, sleeping and retrying", err)
+			workflow.Sleep(ctx, params.PollingInterval)
```
cretz, line 33: This can error

cretz (commented):
Looks mostly good, just some minor cleanup needed

polling/testservice.go
```diff
@@ -0,0 +1,25 @@
+package polling
```
cretz, line 1: Can you add a README file in this top-level `polling` dir that links to the three child samples? Also, please update root README to point to this sample dir.

polling/frequent/activities.go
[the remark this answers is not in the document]
edmondop: Handling this now, apologies. I'll use this thread to learn something about GoLang. 

how is time.After() different than time.sleep()?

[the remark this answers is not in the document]
cretz: The former returns a channel that can be awaited on with another channel (such as checking context done at the same time), the latter is a normal blocking call. See https://pkg.go.dev/context and https://go.dev/blog/context to learn more about contexts.

polling/infrequent/worker/main.go
```diff
@@ -0,0 +1,36 @@
+package main
+
+import (
+	"github.com/temporalio/samples-go/polling"
```
cretz, line 4: These imports are all still out of order

polling/frequent/activities.go
```diff
@@ -0,0 +1,33 @@
...
+	TestService  *polling.TestService
+	PollInterval time.Duration
+}
+
+// DoPoll Activity.
+// In this activity polling is implemented within the activity itself and not the workflow,
+// using the heartbeat mechanism to keep the activity alive
+func (a *PollingActivities) DoPoll(ctx context.Context) (string, error) {
+	for {
+		res, err := a.TestService.GetServiceResult(ctx)
```
cretz, line 21: Mentioned in another comment, but this is the problem with using errors for actual errors and for not-done-yet poll responses. This error is swallowed. What if a serious error is happening when getting service result? There are no logs. But you probably don't want to log the not-done-yet error, just serious errors (and maybe return them so activity retry mechanism can kick in).

polling/periodic_sequence/child_workflow.go
```diff
@@ -0,0 +1,46 @@
...
+		if err == nil {
+			return pollResult, nil
+		} else {
+			logger.Error("Error in activity, sleeping and retrying", err)
+			err := workflow.Sleep(ctx, params.PollingInterval)
+			if err != nil {
+				return "", err
+			}
+
+		}
```
cretz, line 41:
```suggestion
		}
		logger.Error("Error in activity, sleeping and retrying", err)
		err := workflow.Sleep(ctx, params.PollingInterval)
		if err != nil {
			return "", err
		}
```

polling/testservice.go
```diff
@@ -0,0 +1,26 @@
...
+}
+
+func NewTestService(errorAttempts int) TestService {
+	return TestService{
+		tryAttempts:   0,
+		errorAttempts: errorAttempts,
+	}
+}
+
+func (testService *TestService) GetServiceResult(ctx context.Context) (string, error) {
```
cretz, line 20: Mentioned above, but I would differentiate poll fail from service fail for clarity. Not required but would be clearer.

polling/frequent/starter/main.go
```diff
@@ -0,0 +1,32 @@
+package main
+
+import (
```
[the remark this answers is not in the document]
edmondop: Is this still applicable @cretz ? Does gofmt also execute goimports?

polling/testservice.go
edmondop: How is poll failing possible here? I am not sure I understood

polling/infrequent/activities.go
```diff
@@ -0,0 +1,15 @@
...
+	"github.com/temporalio/samples-go/polling"
+)
+
+type PollingActivities struct {
+	TestService *polling.TestService
+}
+
+// DoPoll Activity.
+func (a *PollingActivities) DoPoll(cmd context.Context) (string, error) {
+	return a.TestService.GetServiceResult(cmd)
```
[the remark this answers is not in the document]
edmondop: Should the comment be here or in the place where I do the retry? I was thinking something like

// Assuming the failure is intermittent and retrying after some time

polling/frequent/starter/main.go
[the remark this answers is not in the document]
cretz: I believe `go fmt` does the right thing. But please make sure it is executed on all files.

polling/infrequent/activities.go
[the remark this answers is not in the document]
cretz: Anywhere, and maybe on the `frequent` example saying we are intentionally hiding errors

polling/infrequent/worker/main.go
```diff
@@ -0,0 +1,37 @@
+package main
+
+import (
+	"log"
+
+	"github.com/temporalio/samples-go/polling"
```
cretz, line 6: No need for blank lines between the non-stdlib imports. This should be grouped with the other one further down IMO.

edmondop: Interesting, `goimports -w .` do not fix the issue neither does running go Imports from GoLand.

cretz (approved):

PR #253, opened 2023-02-14
Add an update sample

mmcshane:
## What was changed

Add an update sample

## Why?

To demonstrate update functionality

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/update/starter/main.go
+++ b/update/starter/main.go
@@ -1,0 +1,55 @@
+package main
+
+import (
+	"context"
+	"log"
+
+	"github.com/temporalio/samples-go/update"
+	"go.temporal.io/sdk/client"
+)
+
+func main() {
+	c, err := client.Dial(client.Options{})
+	if err != nil {
+		log.Fatalln("Unable to create client", err)
+	}
+	defer c.Close()
+
+	workflowOptions := client.StartWorkflowOptions{
+		ID:        "update-workflow-ID",
+		TaskQueue: "update",
+	}
+
+	we, err := c.ExecuteWorkflow(context.Background(), workflowOptions, update.Counter)
+	if err != nil {
+		log.Fatalln("Unable to execute workflow", err)
+	}
+
+	log.Println("Started workflow", "WorkflowID", we.GetID(), "RunID", we.GetRunID())
+
+	mult := 1
+	for i := 0; i < 10; i++ {
+		addend := mult * i
+		mult *= -1 // flip addend between negative and positive for each iteration
+		handle, err := c.UpdateWorkflow(context.Background(), we.GetID(), we.GetRunID(), update.FetchAndAdd, addend)
+		if err != nil {
+			log.Fatal("error issuing update request", err)
+		}
+		var result int
+		err = handle.Get(context.Background(), &result)
+		if err != nil {
+			log.Printf("fetch_and_add with addend %v failed: %v", addend, err)
+		} else {
+			log.Printf("fetch_and_add with addend %v succeeded: %v", addend, result)
+		}
+	}
+
+	if err = c.SignalWorkflow(context.Background(), we.GetID(), we.GetRunID(), update.Done, nil); err != nil {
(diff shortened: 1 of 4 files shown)


update/starter/main.go
```diff
@@ -0,0 +1,58 @@
...
+	}
+
+	we, err := c.ExecuteWorkflow(context.Background(), workflowOptions, update.Counter)
+	if err != nil {
+		log.Fatalln("Unable to execute workflow", err)
+	}
+
+	log.Println("Started workflow", "WorkflowID", we.GetID(), "RunID", we.GetRunID())
+
+	mult := 1
```
cretz, line 30: Might as well throw a comment above this saying what it's doing with these values (can of course be small, because README and workflow itself kinda show what's happening)

update/starter/main.go
```diff
@@ -0,0 +1,58 @@
...
+	for i := 0; i < 10; i++ {
+		addend := mult * i
+		mult *= -1
+		handle, err := c.UpdateWorkflow(context.Background(), we.GetID(), we.GetRunID(), "fetch_and_add", addend)
+		if err != nil {
+			log.Fatal("error issuing update request", err)
```
cretz, line 36: Inconsistent capitalization on logs

update/update.go
```diff
@@ -0,0 +1,36 @@
...
+import (
+	"fmt"
+
+	"go.temporal.io/sdk/workflow"
+)
+
+func Counter(ctx workflow.Context) (int, error) {
+	counter := 0
+
+	_ = workflow.SetUpdateHandlerWithOptions(
```
cretz, line 12: Might as well bubble these errors out to be good example setters

update/starter/main.go
```diff
@@ -0,0 +1,58 @@
...
+		err = handle.Get(context.Background(), &result)
+		if err != nil {
+			log.Printf("fetch_and_add with addend %v failed: %v", addend, err)
+		} else {
+			log.Printf("fetch_and_add with addend %v succeeded: %v", addend, result)
+		}
+	}
+
+	// Synchronously wait for the workflow completion.
+	err = c.CancelWorkflow(context.Background(), we.GetID(), "")
```
cretz, line 48: Kinda pedantic, but would prefer to use a signal if the goal is "stop workflow successfully". Trying not to encourage people to swallow cancellation.

cretz (commented):
Minor comments. Looks great. Be prepared for the onslaught of users (I know I'm never gonna use signals again).

update/starter/main.go
mmcshane: Added on the `*= -1` line.

update/starter/main.go
cretz: The workflow will show as succeeded instead of cancelled when you cancel here because you don't bubble the context error on `_ = ctx.Done().Receive(ctx, nil)`.

It is definitely one of the simplest ways, but definitely not one of the most idiomatic ways to complete a workflow _successfully_. People struggle with swallowing this error because they don't always know when it will appear (it may be as the result of an activity or something happening in that update handler for instance instead of on Done). Yes you happen to not use this context in multiple cancellation places but most do.

We try to encourage people to bubble out cancellation errors wherever they occur instead of using cancellation in place of a signal. Otherwise the context use becomes muddied however its used elsewhere in the workflow. And in other languages it's even more painful to swallow a cancel as if everything's ok (though technically doable).

It's basically just as easy to:

```suggestion
	err = c.SignalWorkflow(context.Background(), we.GetID(), "", "finish-workflow", "")
```

And change `_ = ctx.Done().Receive(ctx, nil)` to `_ = workflow.GetSignalChannel(ctx, "finish-workflow").Receive(ctx, nil)`. And if you wanted cancellation you could add a selector of course, but at least you haven't written a workflow that can never appear cancelled by co-opting cancellation to mean successfully complete.

(again, more of a best practices thing than a whether-it-works thing)

cretz (approved):

PR #254, opened 2023-02-17
Update CI Go version to 1.20

Quinn-With-Two-Ns:
Update CI to latest and supported version of golang


Quinn-With-Two-Ns:
need to update the Go SDK to pickup the fixed workflow checker

PR #256, opened 2023-02-23
bump images in docker compose for Gitpod users

taoza:
## What was changed

- Ported the latest https://github.com/temporalio/docker-compose
- Refactored .gitpod.yml to wait for docker-compose's bootstrap more gracefully

## Why?

As also reported on [slack](https://temporalio.slack.com/archives/CTDTU3J4T/p1677123780129219), some of the example code with newer features, such as [Memo](https://github.com/temporalio/samples-go/tree/main/memo), are failing on Gitpod due to some rather old images referenced in docker-compose.yml.

This PR fixes the issue by porting the latest docker-compose config from https://github.com/temporalio/docker-compose, including the versioning `.env` file and the `dynamicconfig`.

## Question

Shall we setup some automation to avoid this kind of manual sync in the future?

## Checklist

1. Closes 

N/A

3. How was this tested:

- Follow this Gitpod link: [https://gitpod.io/#https://github.com/taoza/samples-go/tree/taoza/bump-images-in-docker-compose](https://gitpod.io/#https://github.com/taoza/samples-go/tree/taoza/bump-images-in-docker-compose)
- Check if the hello world example is successful
- Check if Temporal web UI is loaded properly after waiting for Temporal to be ready
- Make sure the new [Memo example](https://github.com/temporalio/samples-go/tree/main/memo) runs correctly

4. Any docs updates needed?

No.


bergundy (approved):
Thanks!

We'll soon want to use Temporal CLI instead of this complex compose based environment but it's awesome that we have this fixed.

PR #273, opened 2023-05-01
TW-49: snipsynced smaller code segments

DJSanti:
## What was changed
The starter/main.go file for the schedule sample has been divided into smaller snipsyncs.

## Why?
The code was too large to encapsulate in one snipsync. This should fix a holdup on "How to Schedule a Workflow in Go", where the code wasn't rendering correctly.

## Any docs updates needed?
"How to Schedule a Workflow in Go" will be regenerated to test this.


Quinn-With-Two-Ns (approved):

DJSanti:
Going to merge and see if this fixes it. Patrick also concurs on my issue being based on snippet size

PR #274, opened 2023-05-04
Make server listen to all interfaces instead of just loopback

taoza:
## What was changed

Update grpc-proxy server to listen to all network interfaces.

## Why?

We should not assume request always come from the loopback interface even for local dev environment. For example, if the server runs inside of a container, a request originated from either the host or another container is routed through the VM's interface instead of loopback.

This was originally reported on the forum:
https://community.temporal.io/t/running-temporal-proxy-server-in-docker/8107

### DIFF
--- a/grpc-proxy/README.md
+++ b/grpc-proxy/README.md
@@ -6,15 +6,15 @@ This sample shows how to compress payloads using a GRPC proxy. This allows manag
 1) Run a [Temporal service](https://github.com/temporalio/samples-go/tree/main/#how-to-use).
 2) Run the following command to start the GRPC proxy listening on port 8081
 ```
-go run proxy-server/main.go
+go run grpc-proxy/proxy-server/*.go
 ```
 3) Run the following command to start the worker. The worker is configured to connect to the proxy.
 ```
-go run worker/main.go
+go run grpc-proxy/worker/main.go
 ```
 4) Run the following command to start the example. The client the starter uses is configured to connect to the proxy.
 ```
-go run starter/main.go
+go run grpc-proxy/starter/main.go
 ```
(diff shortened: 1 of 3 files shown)


grpc-proxy/README.md
```diff
@@ -6,15 +6,15 @@ This sample shows how to compress payloads using a GRPC proxy. This allows manag
...
 ```
 3) Run the following command to start the worker. The worker is configured to connect to the proxy.
 ```
-go run worker/main.go
+go run grpc-proxy/worker/main.go
 ```
 4) Run the following command to start the example. The client the starter uses is configured to connect to the proxy.
 ```
-go run starter/main.go
+go run grpc-proxy/starter/main.go
```
taoza, line 17: Just to make sure those commands are consistent with other examples.

cretz:
> We should not assume request always come from the loopback interface even for local dev environment.

But I think we can make these assumptions for our example. The example is not meant to be general purpose, it's meant to be a localhost proxy. Note how all of the other samples assume a localhost Temporal server instead of supporting other non-localhost scenarios. I think we expect people using the sample outside of the sample instructions would alter the sample for their needs.

Having said that, I don't _mind_ changing it, but I think localhost assumption is fine for all of our samples.

taoza:
Thanks @cretz I do agree it's reasonable to have that assumption. But perhaps another way to approach this is to follow the common practice that server don't usually announce interface to listen to unless there is a really good reason.

cretz (approved):

taoza:
> But at this point, it's not a big enough deal, I'll approve. Just make sure you don't encourage people to actually use it anywhere but localhost.

Absolutely @cretz. Thanks for the review!

PR #276, opened 2023-05-11
update cancellation sample

tsurdilo:
Updates cancellation sample:
1) don't swallow error from activity
2) don't swallow error in activity code since activity options define WaitForCancellation: true


cancellation/activity.go
```diff
@@ -21,7 +21,7 @@ func (a *Activities) ActivityToBeCanceled(ctx context.Context) (string, error) {
 			activity.RecordHeartbeat(ctx, "")
 		case <-ctx.Done():
 			logger.Info("context is cancelled")
-			return "I am canceled by Done", nil
+			return "I am canceled by Done", ctx.Err()
```
cretz, line 24: No reason to return a value if you're returning cancellation

cancellation/workflow.go
```diff
@@ -31,10 +31,14 @@ func YourWorkflow(ctx workflow.Context) error {
...
 		}
+
 	}()
 
 	var result string
 	err := workflow.ExecuteActivity(ctx, a.ActivityToBeCanceled).Get(ctx, &result)
+	if err != nil {
+		return err
+	}
 	logger.Info(fmt.Sprintf("ActivityToBeCanceled returns %v, %v", result, err))
```
cretz, line 42: No reason to log an error you know is nil

cancellation/activity.go
tsurdilo: done :)

cancellation/workflow.go
```diff
@@ -31,11 +30,14 @@ func YourWorkflow(ctx workflow.Context) error {
...
 
 	var result string
 	err := workflow.ExecuteActivity(ctx, a.ActivityToBeCanceled).Get(ctx, &result)
-	logger.Info(fmt.Sprintf("ActivityToBeCanceled returns %v, %v", result, err))
+	if err != nil {
+		return err
+	}
 
 	err = workflow.ExecuteActivity(ctx, a.ActivityToBeSkipped).Get(ctx, nil)
 	logger.Error("Error from ActivityToBeSkipped", "Error", err)
```
cretz, line 43: The original purpose of this sample was to show that this is skipped because the context is already cancelled. Now that we return early unlike before, this is essentially unreachable code. Do you want to remove it?

tsurdilo: done, also added cleanup activity specific activity options (as it can fail heartbeat timeout, and its not heartbeating anyways)

cancellation/workflow.go
```diff
@@ -26,21 +25,25 @@ func YourWorkflow(ctx workflow.Context) error {
...
 	err := workflow.ExecuteActivity(ctx, a.ActivityToBeCanceled).Get(ctx, &result)
-	logger.Info(fmt.Sprintf("ActivityToBeCanceled returns %v, %v", result, err))
-
-	err = workflow.ExecuteActivity(ctx, a.ActivityToBeSkipped).Get(ctx, nil)
-	logger.Error("Error from ActivityToBeSkipped", "Error", err)
+	if err != nil {
+		return err
+	}
 
 	logger.Info("Workflow Execution complete.")
```
cretz, line 47: Unreachable code, but in this case I think it's fine

cretz (approved):
May want @Quinn-With-Two-Ns to take a peek also

Quinn-With-Two-Ns (approved):

Quinn-With-Two-Ns:
@tsurdilo did you want to merge this?

PR #281, opened 2023-06-08
Batch sliding window sample

mfateev:
Added a new batch sliding windows sample.


Quinn-With-Two-Ns:
I don't have any major comments on the code itself, just  small nits.

I agree with @cretz that it is difficult to follow. I think it would be clearer if the readme outlined what situation a user should look at this sample and describe how the different workflows interact. I don't think `ProcessBatchWorkflow` is even called out in the readme. 

There are a number of design decisions being made in this sample that lead to additional complexity. Since this is for education I think we should call out why we are taking on this complexity. If I understand the sample correctly it seems like we are trying to show two different things:

1. Using `ProcessBatchWorkflow` to help with throughput? though I don't think we explain how this helps with throughput in the sample

2. Using `ChildWorkflows` and signals with `SlidingWindowWorkflow`  to _continue as new_ while processing records instead of a more basic  like just not processing new records until we can _continue as new_.

Maybe it would be clearer as two samples one showing `SlidingWindowWorkflow` and one showing  `ProcessBatchWorkflow` that has a much simpler child workflow or depend on the  `SlidingWindowWorkflow`  sample?

mfateev:
In Java, we have three different samples for batch: https://github.com/temporalio/samples-java/tree/main/core/src/main/java/io/temporal/samples/batch. Here I ported the most complex one. It would be nice if someone ported the other two.

Quinn-With-Two-Ns:
Can we link this sample in the [README](https://github.com/temporalio/samples-go#scenario-based-examples)

Quinn-With-Two-Ns (approved):

tsurdilo (approved):

PR #285, opened 2023-06-09
Add build id based versioning samples

Sushisource:
Add samples for build-id-based versioning. 

Closes https://github.com/temporalio/samples-go/issues/284


build-id-versioning/starter/main.go
```diff
@@ -0,0 +1,132 @@
...
+	"github.com/pborman/uuid"
+
+	"go.temporal.io/sdk/client"
+)
+
+func main() {
+	ctx := context.Background()
+
+	// Get task queue name from CLI arg
+	taskQueue := os.Args[1]
```
cretz, line 18: Most of our samples just use a hardcoded task queue and don't require CLI input, I think this can do that too

build-id-versioning/starter/main.go
```diff
@@ -0,0 +1,132 @@
...
+	if taskQueue == "" {
+		log.Fatalln("Must provide task queue name as first and only argument")
+	}
+	log.Println("Using task queue", taskQueue)
+
+	// The client is a heavyweight object that should be created once per process.
+	c, err := client.Dial(client.Options{
+		HostPort: client.DefaultHostPort,
+	})
```
cretz, line 27:
```suggestion
	c, err := client.Dial(client.Options{})
```

(sorry, I know we probably have this default elsewhere and this was just pasted from there, don't have to fix)

build-id-versioning/starter/main.go
```diff
@@ -0,0 +1,132 @@
...
+	if err != nil {
+		log.Fatalln("Unable to create client", err)
+	}
+	defer c.Close()
+
+	// First, let's make the task queue use the build id versioning feature by adding an initial
+	// default version to the queue:
+	err = c.UpdateWorkerBuildIdCompatibility(ctx, &client.UpdateWorkerBuildIdCompatibilityOptions{
+		TaskQueue: taskQueue,
+		Operation: &client.BuildIDOpAddNewIDInNewDefaultSet{
```
cretz, line 37: That is a mouthful, heh. Would have supported something like `BuildIDOpAddSet`, but this was good enough for the SDK review so good enough here. It's not very Go like to usually have these full-sentence identifiers.

build-id-versioning/worker/main.go
```diff
@@ -0,0 +1,56 @@
...
+
+	// We will start a handful of workers, each with a different build identifier.
+	wg := sync.WaitGroup{}
+	createAndRunWorker(c, taskQueue, "1.0", build_id_versioning.SampleChangingWorkflowV1, &wg)
+	createAndRunWorker(c, taskQueue, "1.1", build_id_versioning.SampleChangingWorkflowV1b, &wg)
+	createAndRunWorker(c, taskQueue, "2.0", build_id_versioning.SampleChangingWorkflowV2, &wg)
+	wg.Wait()
+}
+
+func createAndRunWorker(c client.Client, taskQueue, buildID string, workflowFunc interface{}, wg *sync.WaitGroup) {
```
cretz, line 35: Can take a callback to not harm the static analyzer here

```suggestion
func createAndRunWorker(c client.Client, taskQueue, buildID string, wg *sync.WaitGroup, register func(w *worker.Worker)) {
```

And then just call as:

```go
createAndRunWorker(c, taskQueue, "1.0", &wg, func (w *worker.Worker) {
    w.RegisterWorkflowWithOptions(
        build_id_versioning.SampleChangingWorkflowV1, 
        workflow.RegisterOptions{Name: "SampleChangingWorkflow"},
    )
})
```

build-id-versioning/workflows.go
```diff
@@ -0,0 +1,134 @@
...
+	for {
+		var signal string
+		signalChan.Receive(ctx, &signal)
+
+		if signal == "do-activity" {
+			workflow.GetLogger(ctx).Info("Changing workflow v1 running activity")
+			ao := workflow.ActivityOptions{
+				StartToCloseTimeout: 10 * time.Second,
+			}
+			ctx1 := workflow.WithActivityOptions(ctx, ao)
```
cretz, line 27: Feel free to move out of loop near top if you want (or at least use the created context to also `Get`, though not required is often best practice). Will definitely cut down on repeated code in your next workflow.

build-id-versioning/workflows.go
```diff
@@ -0,0 +1,134 @@
...
+			if err != nil {
+				return err
+			}
+		} else {
+			workflow.GetLogger(ctx).Info("Concluding workflow v1", workflow.Now(ctx))
+			break
+		}
+	}
+
+	return nil
```
cretz, line 38:
```suggestion
			return nil
		}
	}
```

build-id-versioning/workflows.go
```diff
@@ -0,0 +1,134 @@
...
+			ao := workflow.ActivityOptions{
+				StartToCloseTimeout: 10 * time.Second,
+			}
+			ctx1 := workflow.WithActivityOptions(ctx, ao)
+			err := workflow.ExecuteActivity(ctx1, SomeActivity, "v1").Get(ctx, nil)
+			if err != nil {
+				return err
+			}
+		} else {
+			workflow.GetLogger(ctx).Info("Concluding workflow v1", workflow.Now(ctx))
```
cretz, line 33: Loggers expect key-value contextual data (this is missing key). Also note that most users' implementations of loggers put the time for them

build-id-versioning/starter/main.go
Sushisource: It needs to start with an unversioned queue every time to work. The alternative would be moving the starter and workers into one file.

build-id-versioning/worker/main.go
Sushisource: I merged the fix I needed anyway. Doing this takes away part of the point which is that the same name always gets used

build-id-versioning/starter/main.go
bergundy: I kinda like that we're putting the HostPort in the samples, it teaches where this option can be provided and what the default is.

build-id-versioning/starter/main.go
```diff
@@ -0,0 +1,132 @@
...
+
+	// Give a chance for these signals to be processed by the 1.0 worker
+	time.Sleep(5 * time.Second)
+
+	// Now, let's update the task queue with a new compatible version:
+	err = c.UpdateWorkerBuildIdCompatibility(ctx, &client.UpdateWorkerBuildIdCompatibilityOptions{
+		TaskQueue: taskQueue,
+		Operation: &client.BuildIDOpAddNewCompatibleVersion{
+			BuildID:                   "1.1",
+			ExistingCompatibleBuildId: "1.0",
```
bergundy, line 75: We missed this inconsistency in the Go SDK.

build-id-versioning/worker/main.go
bergundy: Why is this harming the static analyzer?

Sushisource: It can't follow higher-order function refs to go analyze them, but, it's not exactly important that it does here.

build-id-versioning/workflows.go
Sushisource: I think it's worth keeping just to preserve the idea that you want the exact same code in the old branch.

build-id-versioning/starter/main.go
Sushisource: We can change it if we want, not the most noticeable

build-id-versioning/starter/main.go
cretz: I think `"build-id-versioning-" + uuid.New()` in the worker can just be a fixed value too

build-id-versioning/starter/main.go
cretz: But this doesn't show what the default is, just puts a constant reference. Are you saying we should put `"localhost:7233"` here and in future examples for clarity reasons? Doesn't matter much either way to me.

build-id-versioning/worker/main.go
cretz: Technically we could make it do so (we would taint any expression that _becomes_ a workflow function as needing to be deterministic, and then we'd know this param would need to be), but it's not the simplest or that common.

build-id-versioning/starter/main.go
Sushisource: That means re-running the sample breaks - it needs to be a new TQ each time.

build-id-versioning/starter/main.go
bergundy: I mean it's a "go to definition" (or 2 with the whole internal situation) away but in any case this is insignificant.

build-id-versioning/starter/main.go
cretz: > That means re-running the sample breaks

This is the case with many of our samples

cretz (approved):

build-id-versioning/README.md
```diff
@@ -0,0 +1,23 @@
+# Build ID Based Versioning
```
cretz, line 1: I wonder if "worker versioning" is a better term

PR #286, opened 2023-06-11
Async activity Invocation Example in Go

edmondop:
## What was changed

This adds an example of a non-blocking invocation of two activities in Go, with a behavior coherent with the Java example here
https://github.com/edmondop/samples-java/blob/main/src/main/java/io/temporal/samples/hello/HelloAsync.java#L114

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/README.md
+++ b/README.md
@@ -222,7 +222,6 @@ resource waiting its successful completion
 
 Mostly examples we haven't yet ported from https://github.com/temporalio/samples-java/
 
-- Async activity calling: *Example to be completed*
 - Async lambda:  *Example to be completed*
 - Periodic Workflow: Workflow that executes some logic periodically. *Example to be completed*
 - Exception propagation and wrapping: *Example to be completed*
--- a/activities-async/workflow.go
+++ b/activities-async/workflow.go
@@ -1,0 +1,30 @@
+package activities_async
+
+import (
+	"go.temporal.io/sdk/workflow"
+	"time"
+)
+
+func AsyncActivitiesWorkflow(ctx workflow.Context) (res string, err error) {
+	selector := workflow.NewSelector(ctx)
+	var res1 string
+	var res2 string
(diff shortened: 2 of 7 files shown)


README.md
cretz, line 1: Add this example to the example list

activities-async/workflow.go
```diff
@@ -0,0 +1,30 @@
...
+	fut1 := workflow.ExecuteActivity(ctx, SayHello, "Temporal")
+	fut2 := workflow.ExecuteActivity(ctx, SayGoodbye, "Temporal")
+	selector.AddFuture(fut1, func(future workflow.Future) {
+		err = future.Get(ctx, &res1)
+	})
+	selector.AddFuture(fut2, func(future workflow.Future) {
+		err = future.Get(ctx, &res2)
+	})
+	selector.Select(ctx)
+	selector.Select(ctx)
```
cretz, line 25: If an `err` was reported by the first select, you have swallowed it

README.md
edmondop: Fixed with https://github.com/temporalio/samples-go/pull/286/commits/a063d6cc5c260b350c6d3090a898a7517598f1cf

activities-async/workflow.go
edmondop: Fixed with https://github.com/temporalio/samples-go/pull/286/commits/7156e315a7a30177366bdb2c5b692f783b789894

cretz (approved):

edmondop:
> This use case is already covered by existing samples:
> 
> https://github.com/temporalio/samples-go/tree/main/splitmerge-selector https://github.com/temporalio/samples-go/tree/main/splitmerge-future

Thanks! Didn't know those examples! If this pattern is what match more closely the Java async examples, maybe we can just update the README ?

PR #292, opened 2023-07-04
Add update request response sample

Quinn-With-Two-Ns:
Add a samples showing how to sue update for request/response pairs. I tried to match the structure of the existing samples to help users compare the old approaches with this new one.

Note: Since update is not released I held off fully recommending the update approach.

### DIFF
--- a/reqrespupdate/README.md
+++ b/reqrespupdate/README.md
@@ -1,0 +1,49 @@
+# Request/Response Sample with Update-Based Responses
+
+This sample demonstrates how to send a request and get a response from a Temporal workflow via an update.
+
+[Update](https://docs.temporal.io/workflows#update) is a new feature available for preview on [Temporal Server v1.21](https://github.com/temporalio/temporal/releases/tag/v1.21.0).
+
+### Running
+
+Follow the below steps to run this sample:
+
+1) Run a [Temporal service](https://github.com/temporalio/samples-go/tree/main/#how-to-use). Note, update is only supported on [Temporal Server v1.21](https://github.com/temporalio/temporal/releases/tag/v1.21.0)
+
+2) Update functionality is disabled by default but can be enabled per Namespace by setting the `frontend.enableUpdateWorkflowExecution` flag to true for that Namespace in dynamic config.
+
+3) Run the following command in the background or in another terminal to run the worker:
+
+    go run ./reqrespupdate/worker
+
+4) Run the following command to start the workflow:
+
+    go run ./reqrespupdate/starter
+
+5) Run the following command to uppercase a string every second:
+
+    go run ./reqrespupdate/request
+
+Multiple of those can be run on different terminals to confirm that the processes are independent.
+
+### Comparison with activity-based responses and query-based response
+
+There are two other samlples showing how to do a request response pattern under [reqrespactivity](../reqrespactivity) sample and the [reqrespquery](../reqrespquery). The update based approach is the supperior option, and once released, will be the recommend approach.
+
+
+### Explanation of continue-as-new
+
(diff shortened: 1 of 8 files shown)


reqrespupdate/README.md
```diff
@@ -0,0 +1,49 @@
+# Request/Response Sample with Update-Based Responses
+
+This sample demonstrates how to send a request and get a response from a Temporal workflow via an update.
+
+[Update](https://docs.temporal.io/workflows#update) is a new feature available for preview on [Temporal Server v1.21](https://github.com/temporalio/temporal/releases/tag/v1.21.0).
```
cretz, line 5: May need some internal reminder to come update this once it's out of preview (including some of the verbiage below like "once released")

reqrespupdate/README.md
```diff
@@ -0,0 +1,49 @@
...
+
+5) Run the following command to uppercase a string every second:
+
+    go run ./reqrespupdate/request
+
+Multiple of those can be run on different terminals to confirm that the processes are independent.
+
+### Comparison with activity-based responses and query-based response
+
+There are two other samlples showing how to do a request response pattern under [reqrespactivity](../reqrespactivity) sample and the [reqrespquery](../reqrespquery). The update based approach is the supperior option, and once released, will be the recommend approach.
```
cretz, line 31:
```suggestion
There are two other samples showing how to do a request response pattern under [reqrespactivity](../reqrespactivity) sample and the [reqrespquery](../reqrespquery). The update based approach is the superior option, and once released, will be the recommend approach.
```

reqrespupdate/README.md
```diff
@@ -0,0 +1,49 @@
...
+
+
+### Explanation of continue-as-new
+
+Workflows cannot have infinitely-sized history and when the event count grows too large, `ContinueAsNew` can be returned
+to start a new one atomically. However, in order not to lose any data, update requests must be handles and any other futures
```
cretz, line 37:
```suggestion
to start a new one atomically. However, in order not to lose any data, update requests must be handled and any other futures
```

reqrespupdate/workflow.go
```diff
@@ -0,0 +1,119 @@
...
+	UpdateHandler = "update"
+)
+
+var (
+	ErrBackoff = errors.New("trying to continue as new")
+)
+
+// UppercaseWorkflow is a workflow that accepts requests to uppercase strings
+// via updates and provides a response.
+func UppercaseWorkflow(ctx workflow.Context, rejectUpdateOnPendingContinueAsNew bool) error {
```
cretz, line 24: Would consider `UppercaseWorkflowOptions` struct here just so that what the bool value means caller-side is clearer when set as a field (not required, no biggie)

reqrespupdate/workflow.go
```diff
@@ -0,0 +1,119 @@
...
+	return u, nil
+}
+
+func (u *uppercaser) run(ctx workflow.Context) error {
+	var requestCount int
+	var pendingUpdates int
+
+	var options workflow.UpdateHandlerOptions
+	if u.rejectUpdateOnPendingContinueAsNew {
+		options.Validator = func(ctx workflow.Context, request Request) error {
```
cretz, line 86: This is a great use of validators, may be worth a small comment above clarifying that this doesn't even persist the update in history

reqrespupdate/README.md
cretz, line 1: Can update those other `reqresp` samples' READMEs to reference this one and vice versa, or can wait until updates are more stable, your choice.

cretz (approved):
Nice. No blocking comments.

reqrespupdate/README.md
Quinn-With-Two-Ns: Yeah I purposefully left it off till update is more stable

PR #294, opened 2023-07-08
Allow to build a single sample

mindaugasrukas:
## What was changed
Added option to build a single sample

## Why?
Improving developer experience.

## Checklist

1. Closes 
No

2. How was this tested:
```
% make bins
Build sample for .
go build -o ./bin/./cancellation/cancel ././cancellation/cancel
go build -o ./bin/./cancellation/starter ././cancellation/starter
go build -o ./bin/./cancellation/worker ././cancellation/worker
go build -o ./bin/./encryption/codec-server ././encryption/codec-server
go build -o ./bin/./encryption/starter ././encryption/starter
go build -o ./bin/./encryption/worker ././encryption/worker
go build -o ./bin/./metrics/starter ././metrics/starter
...

% make bins helloworld
Build sample for helloworld
go build -o ./bin/helloworld/starter ./helloworld/starter
go build -o ./bin/helloworld/worker ./helloworld/worker
make: `helloworld' is up to date.
```

3. Any docs updates needed?
Added

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,6 +30,20 @@ Each sample demonstrates one feature of the SDK, together with tests.
 - [**Basic mTLS hello world**](./helloworldmtls): Simple example of a
   Workflow Definition and an Activity Definition using mTLS like Temporal Cloud.
 
+## Building sampes
+
+To build all samples run this command:
+
+  $ make bins
+
+To build only a single sample run this command:
+
+  $ make bins <example-name>
+
+For example:
+
+  $ make bins helloworld
+
 ### API demonstrations
 
 - **Async activity completion**: Example of
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@ Each sample demonstrates one feature of the SDK, together with tests.
 - [**Basic mTLS hello world**](./helloworldmtls): Simple example of a
   Workflow Definition and an Activity Definition using mTLS like Temporal Cloud.
 
-## Building sampes
+## Building samples
 
 To build all samples run this command:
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -30,6 +30,20 @@ Each sample demonstrates one feature of the SDK, together with tests.
 - [**Basic mTLS hello world**](./helloworldmtls): Simple example of a
   Workflow Definition and an Activity Definition using mTLS like Temporal Cloud.
 
+## Building sampes
```
cretz, line 33: I don't think we want to promote platform-specific build steps. In fact, we should probably go the other direction and remove the Makefile altogether.

cretz, line 33: What is the use case to build samples instead of just use the instructions in their READMEs to run them?

cretz: > I don't care much as long as there is a solution. What solution do you suggest instead of Makefile? Mage?

Just use `go run`

> My reasoning is that I wanted to build/run only a single sample

See the README for whatever sample you're talking about which will tell you how to run the sample

cretz: > I don't like following the instruction in the README

This is a bit rough. We need to be able to tell people how to run samples. We can't expect people to not read the README. How do you know what parameters to pass to the executable? How do you know that you may need multiple executables for the sample to work?

> I'd like to have a tool instead, like Make, Mage, or similar.

You do have a tool: `go`. In fact, you're typing _more_ characters and maybe making more mistakes by using `make` to just run `go build` for you followed by running the executable. Just `go run`.

PR #297, opened 2023-07-18
Show a way to implement cancellable updates

Quinn-With-Two-Ns:
Show how one could implement cancellable updates purely in user side code.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,6 +4,9 @@ go 1.16
 
 replace github.com/cactus/go-statsd-client => github.com/cactus/go-statsd-client v3.2.1+incompatible
 
+replace (
+	go.temporal.io/sdk v1.23.1 => ../sdk-go
+)
 require (
 	github.com/HdrHistogram/hdrhistogram-go v1.1.2 // indirect
 	github.com/golang/mock v1.6.0
--- a/update-cancel/interceptor.go
+++ b/update-cancel/interceptor.go
@@ -1,0 +1,66 @@
+package update_cancel
+
+import (
+	"errors"
+
+	"go.temporal.io/sdk/interceptor"
+	"go.temporal.io/sdk/workflow"
+)
+
+const (
+	UpdateCancelHandle = "update-cancel"
+)
+
+type workerInterceptor struct {
+	interceptor.WorkerInterceptorBase
+}
+
+func NewWorkerInterceptor() interceptor.WorkerInterceptor {
+	return &workerInterceptor{}
+}
+
+func (w *workerInterceptor) InterceptWorkflow(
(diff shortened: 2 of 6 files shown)


go.mod
```diff
@@ -4,6 +4,9 @@ go 1.16
 
 replace github.com/cactus/go-statsd-client => github.com/cactus/go-statsd-client v3.2.1+incompatible
 
+replace (
+	go.temporal.io/sdk v1.23.1 => ../sdk-go
```
Quinn-With-Two-Ns, line 8: Update interceptors don't work on the current release of the Go SDK

update-cancel/interceptor.go
cretz, line 1: Hrmm, not sure this is a pattern we should encourage, especially hiding this logic in an interceptor. I think it may be clearer if we should how to cancel an update context from another update explicitly in workflow code. If they want to make that generic to apply to all updates hidden away in an interceptor, then they can.

Quinn-With-Two-Ns: The reason I moved it into the interceptor is to show you don't need to explicitly add support in your workflow

Quinn-With-Two-Ns: it just makes the code a less elegant to do it in the workflow, but for the samples repo your right we shouldn't encourage usage of interceptors like this

cretz: Definitely less elegant and less reusable. Acknowledged. Kinda on purpose, which is admittedly a bit strange, but we don't want people to blindly copy the approach. What I've found is that many times people want to customize these patterns so seeing them in-workflow is easy to understand.

And really, if we wanted to write this properly, I'd recommend a utility instead of an interceptor

update-cancel/interceptor.go
```diff
@@ -0,0 +1,66 @@
...
+
+func (w *workflowInboundInterceptor) Init(outbound interceptor.WorkflowOutboundInterceptor) error {
+	w.ctxMap = make(map[string]workflow.CancelFunc)
+	return w.Next.Init(outbound)
+}
+
+func (w *workflowInboundInterceptor) ExecuteWorkflow(ctx workflow.Context, in *interceptor.ExecuteWorkflowInput) (interface{}, error) {
+	err := workflow.SetUpdateHandlerWithOptions(ctx, UpdateCancelHandle, func(ctx workflow.Context, updateID string) error {
+		// Cancel the update
+		w.ctxMap[updateID]()
```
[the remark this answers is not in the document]
Quinn-With-Two-Ns: I didn't delete because I wanted duplicate cancels to behave the same as the server, but I didn't implement that behaviour yet

PR #303, opened 2023-09-18
Add session failure sample

Quinn-With-Two-Ns:
Add a sample showing how to handle a session failure in a workflow since it is not obvious and some users have misconceptions about how failure is handled in a session.

### DIFF
--- a/session-failure/workflow.go
+++ b/session-failure/workflow.go
@@ -1,0 +1,91 @@
+package fileprocessing
+
+import (
+	"errors"
+	"time"
+
+	"go.temporal.io/sdk/temporal"
+	"go.temporal.io/sdk/workflow"
+)
+
+var (
+	ErrSessionHostDown = errors.New("session host down")
+)
+
+// SampleSessionFailureRecoveryWorkflow workflow definition
+func SampleSessionFailureRecoveryWorkflow(ctx workflow.Context, fileName string) (err error) {
+
+	err = runSession(ctx, fileName)
+	numOfRetries := 10
+	for err != nil && numOfRetries >= 0 {
+		// Only retry if we detected the session failed. In a production application
+		// it may make sense to also retry if some other errors occur, it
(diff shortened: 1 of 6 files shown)


session-failure/workflow.go
```diff
@@ -0,0 +1,91 @@
...
+)
+
+var (
+	ErrSessionHostDown = errors.New("session host down")
+)
+
+// SampleSessionFailureRecoveryWorkflow workflow definition
+func SampleSessionFailureRecoveryWorkflow(ctx workflow.Context, fileName string) (err error) {
+
+	err = runSession(ctx, fileName)
```
cretz, line 18:
```go
for retryNum := 0; retryNum < 10; retryNum++ {
    if err = runSession(ctx, fileName); errors.Is(err, ErrSessionDown) {
        if sleepErr := workflow.Sleep(ctx, 5 * time.Minute); sleepErr != nil {
            return sleepErr
        }
        continue
    }
    if err != nil {
        workflow.GetLogger(ctx).Error("Workflow failed.", "Error", err.Error())
    } else {
        workflow.GetLogger(ctx).Info("Workflow completed.")
    }
    return
}
workflow.GetLogger(ctx).Error("Workflow failed after multiple session retries.", "Error", err.Error())
return
```

Reads a bit clearer to me, but no change needed

session-failure/workflow.go
```diff
@@ -0,0 +1,91 @@
...
+	numOfRetries := 10
+	for err != nil && numOfRetries >= 0 {
+		// Only retry if we detected the session failed. In a production application
+		// it may make sense to also retry if some other errors occur, it
+		// depends on your business logic.
+		if errors.Is(err, ErrSessionHostDown) {
+			workflow.Sleep(ctx, 5*time.Minute)
```
cretz, line 25: React to error from sleep

cretz (approved):

Issue #301, opened 2023-08-31
[Bug] serverjwtauth sample doesn't work with newer servers


cretz:
### Describe the bug

Multiple server changes were made since this sample was written. Specifically we need to:

* Update README to have them set `USE_INTERNAL_FRONTEND=true` env var and `SERVICES=frontend:internal-frontend:history:matching:worker`
* Update https://github.com/temporalio/samples-go/blob/6b93fe45a8824a56ac70f30832d262b94921a80d/serverjwtauth/key/main.go#L95C35-L95C35 to change `system:admin` to `temporal-system:admin`

PR #307, opened 2023-10-08
fix temporal server start up error

coolyrat:
## What was changed

Update the serverjwtauth sample to compatible with the latest temporal server version

## Why?

Before making the change, docker-compose up the temporal server error when start. And the README.md is not updated after [v1.20.0](https://github.com/temporalio/temporal/releases/tag/v1.20.0)

## Checklist

1. Closes 

2. How was this tested:

After the change, can run the sample successfully

3. Any docs updates needed?

Updated README.md to help others from having the same problem


dnr:
Note that this fixes #301

cretz (approved):

cretz:
Thanks! Sorry it took so long to get this in.

PR #319, opened 2023-12-27
Add a Datadog sample

Quinn-With-Two-Ns:
Add a sample showing how to use the Datadog interceptor and also point out how to setup metrics and logging


Quinn-With-Two-Ns:
related https://github.com/temporalio/features/issues/380

Quinn-With-Two-Ns:
This RP includes changes from https://github.com/temporalio/samples-go/pull/325 to make the dependencies happy

cretz (approved):

PR #296, opened 2023-07-13
OpenTelemetry instrumentation example

emanuelef:
## What was changed
Added example to show how to instrument OpenTelemetry in the worker and client

## Why?

Currently there is no sample to show how to instrument a Temporal worker and a client with OpenTelemetry.  
This can save some time and clarifies what is the minimum needed to be able to export traces.

There's also an open issue:
https://github.com/temporalio/samples-go/issues/269

## Checklist

1. Closes 

2. How was this tested:

Run as most of the samples, no need to set OTLP endpoint and see the spans in the stdout.  
Also tested with Honeycomb.io.

3. Any docs updates needed?

Updated README.md

PR #293, opened 2023-07-05
Update Temporal server version to v1.22

mindaugasrukas:
## What was changed
Update Temporal server version to v1.22

## Why?
Use the latest release

## Checklist

1. Closes 
No
2. How was this tested:
CI

3. Any docs updates needed?
No

PR #320, opened 2023-12-29
Add open telemetry sample

Quinn-With-Two-Ns:
Add open telemetry sample

Leaving as draft for two reasons:
* Due to some samples depending on an old server version I couldn't add newer otel SDKs so I temporarily removed them
* The unit test is not getting the correct spans and I suspect their is a bug with interceptors in the test enviorment


emanuelef:
Hi, I have a similar problem with my PR: https://github.com/temporalio/samples-go/pull/296
That PR has been there for months but cannot be merged because of https://github.com/temporalio/samples-go/pull/293

Quinn-With-Two-Ns:
@emanuelef Sorry I completely missed your PR or that it was blocked. I will create a PR to upgrade the server dependency.

emanuelef:
@Quinn-With-Two-Ns I see as part of this PR some samples have been removed, did you need that to be able to use OpenTelemetry modules ? I might need to do the same in my PR, provided it makes sense to have both PRs been merged.

emanuelef:
The imports where I'm having problems as well are:

In grpc-proxy and codec-server:
go.temporal.io/server/common/authorization
go.temporal.io/server/common/config

What would be the new way to import those modules ?

emanuelef:
@Quinn-With-Two-Ns The PR https://github.com/temporalio/samples-go/pull/296 is ready to be merged

Quinn-With-Two-Ns:
@emanuelef Thanks, sorry was out of the office.

PR #321, opened 2023-12-31
Update server to v1.22.3

Quinn-With-Two-Ns:
Update Temporal server version to v1.22.3


cretz (approved):
Would also support adding separate `go.mod`s for the few samples that need server/tctl

PR #324, opened 2024-01-17
chore(docs): Update context propagation docs link

gabizou:
Previous link lead to the docs main page, not the intended observability page.


Quinn-With-Two-Ns (approved):
Thanks for the update

Issue #237, opened 2022-12-12
Remove go.temporal.io/server as a dependency


cretz:
### Describe the solution you'd like

Everywhere it's used, `go.temporal.io/server` should be able to be replaced, and where it can't be replaced, we should consider making a separate `go.mod` in that sample. It's a large, moving dependency and an SDK sample demonstrating that a user needs the server as a library indicates a more fundamental problem we should address.

PR #325, opened 2024-01-21
Move codec-server and grpc-proxy to own module

Quinn-With-Two-Ns:
Move codec-server and grpc-proxy to own module. This will let us decouple most of our samples dependencies from the sever dependencies.


Quinn-With-Two-Ns:
closes https://github.com/temporalio/samples-go/issues/237

codec-server/codec-server/codec-server
cretz, line 1: Accidental committed binary?

codec-server/go.mod
```diff
@@ -0,0 +1,87 @@
+module github.com/temporalio/samples-go/codec-server
+
+go 1.21.1
+
+require (
+	github.com/golang/snappy v0.0.4
+	github.com/google/uuid v1.5.0
+	go.temporal.io/api v1.26.0
+	go.temporal.io/sdk v1.25.1
+	go.temporal.io/server v1.22.4
```
cretz, line 10: I don't think the codec server sample should have to depend on the server library at all (other SDKs get along fine without it, though they don't have OIDC, but surely we can copy stuff). But I understand if now is not the time you want to remove this dependency.

codec-server/codec-server/codec-server
Quinn-With-Two-Ns: ah good catch, removed.

codec-server/go.mod
Quinn-With-Two-Ns: Yes I would rather do any changes like that in a separate PR

cretz (approved):

PR #330, opened 2024-02-04
show use of DevServer to test

muralisrini:
## What was changed
Added a test to "hello-world" sample to illustrate use of [DevServer](https://pkg.go.dev/go.temporal.io/sdk/testsuite#DevServer) 

## Why?
Show case DevServer as providing an alternative to testing workflows end-to-end as is without mocks.

## Checklist

1. ran make to verify changes

### DIFF
--- a/helloworld/helloworld_test.go
+++ b/helloworld/helloworld_test.go
@@ -1,12 +1,15 @@
 package helloworld
 
 import (
+	"context"
 	"testing"
 
 	"github.com/stretchr/testify/mock"
-
 	"github.com/stretchr/testify/require"
+
+	"go.temporal.io/sdk/client"
 	"go.temporal.io/sdk/testsuite"
+	"go.temporal.io/sdk/worker"
 )
 
 func Test_Workflow(t *testing.T) {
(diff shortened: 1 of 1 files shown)


helloworld/testserver/env.go
cretz, line 1: Not sure this deserves an entire package of its own. I think the simple single test can manage the dev server and worker lifecycle itself.

helloworld/helloworld_test.go
```diff
@@ -37,3 +39,27 @@ func Test_Activity(t *testing.T) {
...
+	require.NotNil(t, w)
+
+	w.RegisterWorkflow(Workflow)
+	w.RegisterActivity(Activity)
+
+	c, err := testServerEnv.Client()
+	require.NoError(t, err)
+	require.NotNil(t, c)
+
+	result, err := Run(c, testServerEnv.TaskQ)
```
cretz, line 59: For such simple workflow-start code, might as well just do the start from the client in here directly instead of abstracting the simple `ExecuteWorkflow` call as if that part needs testing too. That's just the simplest start-workflow-and-wait, not sure it deserves its own `run.go` file and such, this is a simple hello world. This test can just demonstrate testing the workflow.

helloworld/testserver/env.go
muralisrini: sure. I will remove the package and its abstraction and move the test server code into the test directly.

helloworld/helloworld_test.go
muralisrini: The intent was to abstract the common code and show the actual workflow code (dev, prod etc) can be tested as is with DevServer. You don't think there's value in that ? I can add some comment to the code if that would highlight the intent.

On the other hand, not married to that either if you think duplicating code in test is ok.

helloworld/testserver/env.go
muralisrini: moved server start-up code into the test.

helloworld/helloworld_test.go
cretz: That makes sense to show a client can be used, but I think just using the client to start a workflow in the test the same way the starter's main does is good enough to demonstrate that. My fear though is that we're complicating the hello world just to show this abstraction.

muralisrini: Sure, I'll move that code inline with just a comment stating this is the same code that runs in starter.

muralisrini: Done.

cretz (approved):
Looks great! Will merge if/when CI passes.

PR #332, opened 2024-03-05
Async update sample

Quinn-With-Two-Ns:
Add an example showing async update.

Note: For this sample I used the Go SDK `1.26.0-RC`, but did not update any other samples so CI will fail.


cretz:
(holding off on review pending internal discussions on how start treats wait policy, but can review anyways if you'd like since that only affects some of the comments here)

Issue #333, opened 2024-03-06
[Feature Request] Add Eager Workflow Start sample


antlai-temporal:
### Is your feature request related to a problem? Please describe.
 Add a sample that demonstrates how to start a workflow in eager mode, an experimental latency optimization.

PR #334, opened 2024-03-06
Add eager workflow start sample

antlai-temporal:
## What was changed

Add an example of starting a workflow in eager mode, a latency optimization.
## Why?

We are launching this feature in PreRelease soon

## Checklist

1. Closes 
#333

### DIFF
--- a/eager-workflow-start/README.md
+++ b/eager-workflow-start/README.md
@@ -1,4 +1,13 @@
+### Eager Workflow Start Sample
+
+Eager Workflow Start (EWS) is an experimental latency optimization with the goal of reducing the time it takes to start a workflow.
+
+When the starter and worker are collocated and aware of each other, they can interact while bypassing the server, saving a few time-intensive operations.
+
+Here we modify the `helloworld` sample, ensuring that starter and worker run in the same process, and share the client. Also, the request eager mode flag is set in the start workflow options.
+
 ### Steps to run this sample:
+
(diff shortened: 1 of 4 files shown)


eager-workflow-start/main.go
```diff
@@ -0,0 +1,59 @@
...
+		log.Fatalln("Unable to start worker", err)
+	}
+	defer w.Stop()
+
+	workflowOptions := client.StartWorkflowOptions{
+		ID:        "eager_wf" + uuid.New(),
+		TaskQueue: taskQueueName,
+
+		// 3. Set this flag to true to enable EWS.
+		EnableEagerStart: true,
```
cretz, line 42: Hrmm, wonder if it's a bit much to have an entire sample just to set true on a single workflow option (I wouldn't expect separate samples for every other start option). But I don't mind I guess if it makes sense generally.

Quinn-With-Two-Ns: I think it is worth making a sample for since it also shows you have to reuse the client

Quinn-With-Two-Ns (approved):

eager-workflow-start/README.md
```diff
@@ -0,0 +1,10 @@
+### Steps to run this sample:
```
Quinn-With-Two-Ns, line 1: Can we add a short description here about eager workflow start and how the sample works like some of our other samples?

antlai-temporal: Done

PR #353, opened 2024-06-13
Change link to avoid 404 error

tomwheeler:
## What was changed

I updated the target and text of the link that points to instructions on how to set up Temporal Service for local development. 

## Why?

The old link pointed to a page that no longer exists in the documentation. The text of that link used the title of that page, so it was not appropriate for the page to which it now links. 

## Checklist

1. Closes 

N/A

2. How was this tested:

I verified that the link appeared and functioned as expected in the GitHub Markdown preview. 

3. Any docs updates needed?

No


Quinn-With-Two-Ns (approved):

Quinn-With-Two-Ns:
Thanks Tom!

PR #355, opened 2024-06-14
Adding Accumulator Example

joshmsmith:
## What was changed
Adding accumulator pattern sample: accumulator and associated readme changes

## Why?
This is a pattern that's pretty common, and I wanted to have a sample that I could share about it. Gathering a lot of signals, managing continue as new, demonstrating signal with start, all good stuff to have.

## Checklist

1. Closes N/A

2. How was this tested:
Tested locally, tests in accumulate_signals_workflow_test.go

3. Any docs updates needed?
No


accumulator/accumulate_signals_activities.go
```diff
@@ -0,0 +1,25 @@
...
+	"context"
+	"fmt"
+	"strconv"
+	"go.temporal.io/sdk/activity"
+)
+
+// this activity will process all of the signals together
+func ComposeGreeting(ctx context.Context, s []AccumulateGreeting) (string, error) {
+	log := activity.GetLogger(ctx)
+	if(len(s) < 1) {
```
cretz, line 13: Please make sure `go fmt` is run on all Go files (most editors do it on save by default). This is a bit off as is the lack of blank line separating stdlib imports from non-stdlib. The formatting is off on a lot of code in this PR.

cretz, line 13:
```suggestion
	if len(s) == 0 {
```

Usually a clearer way to check if something is empty

accumulator/accumulate_signals_activities.go
```diff
@@ -0,0 +1,25 @@
...
+		log.Warn("No greetings found when trying to Compose Greetings. ")
```
cretz, line 14:
```suggestion
		log.Warn("No greetings found when trying to Compose Greetings.")
```

Unnecessary trailing space in log

accumulator/accumulate_signals_activities.go
```diff
@@ -0,0 +1,25 @@
...
+	}
+
+	words := "Hello (" + strconv.Itoa(len(s)) + ") Robots"
+	for _, v:= range s {
+		words += fmt.Sprintf(", " + v.GreetingText )
+	}
+
+	words += "!"
```
cretz, line 22:
```suggestion
	words := fmt.Sprintf("Hello (%v) Robots", len(s))
	for _, v := range s {
		words += ", " + v.GreetingText
	}
	words += "!"
```

A bit clearer

accumulator/accumulate_signals_workflow.go
```diff
@@ -0,0 +1,159 @@
...
+ * 
+ * Because Temporal Workflows cannot have an unlimited size, Continue As New is used 
+ *   to process more signals that may come in.
+ * You could create as many groupings as desired, as Temporal Workflows scale out to many workflows without limit.
+ * You could vary the time that the workflow waits for other signals, say for a day, or a variable time from first 
+ *   signal with the GetNextTimeout() function.
+ */
+
+// SignalToSignalTimeout is them maximum time between signals
+var SignalToSignalTimeout = 30 * time.Second
```
cretz, line 29:
```suggestion
const signalToSignalTimeout = 30 * time.Second
```

Use constants if you can. Also no need to expose probably.

accumulator/accumulate_signals_workflow.go
```diff
@@ -0,0 +1,159 @@
...
+
+type AccumulateGreeting struct {
+	GreetingText    string
+	Bucket          string
+	GreetingKey     string
+}
+
+
+// GetNextTimeout returns the maximum time allowed to wait for the next signal.
+func (a *AccumulateGreeting) GetNextTimeout(ctx workflow.Context, timeToExit bool, firstSignalTime time.Time ) (time.Duration, error) {
```
cretz, line 42: No need for this to be public

accumulator/accumulate_signals_workflow.go
```diff
@@ -0,0 +1,159 @@
...
+	}
+	return totalLeft, nil
+}
+
+// AccumulateSignalsWorkflow workflow definition
+func AccumulateSignalsWorkflow(ctx workflow.Context, bucketKey string, greetingsCAN []AccumulateGreeting, uniqueGreetingKeysMapCAN map[string]bool) (allGreetings string, err error) {
+	log := workflow.GetLogger(ctx)
+	var a AccumulateGreeting
+	greetings := []AccumulateGreeting{}
+	greetings = append(greetings, greetingsCAN...)
```
cretz, line 62:
```suggestion
	greetings := greetingsCAN
```

Just use the variables, no need to copy. Also, why "CAN" suffix, why not just accept the values for any reason and it just so happen to be CAN that uses them? You can probably operate on the same workflow state struct throughout instead of moving/copying local variables

accumulator/accumulate_signals_workflow.go
```diff
@@ -0,0 +1,159 @@
...
+		return 0, nil
+	}
+	if SignalToSignalTimeout < totalLeft {
+		return SignalToSignalTimeout, nil
+	}
+	return totalLeft, nil
+}
+
+// AccumulateSignalsWorkflow workflow definition
+func AccumulateSignalsWorkflow(ctx workflow.Context, bucketKey string, greetingsCAN []AccumulateGreeting, uniqueGreetingKeysMapCAN map[string]bool) (allGreetings string, err error) {
```
cretz, line 58: Should make a struct for input as opposed to accepting multiple parameters

accumulator/accumulate_signals_workflow.go
```diff
@@ -0,0 +1,159 @@
...
+	log := workflow.GetLogger(ctx)
+	var a AccumulateGreeting
+	greetings := []AccumulateGreeting{}
+	greetings = append(greetings, greetingsCAN...)
+	unprocessedGreetings := []AccumulateGreeting{}
+	uniqueGreetingKeysMap := make(map[string]bool)
+	for k, v := range uniqueGreetingKeysMapCAN {
+		uniqueGreetingKeysMap[k] = v
+	}
+	var FirstSignalTime time.Time
```
cretz, line 68: Local variables in Go are not capitalized

accumulator/accumulate_signals_workflow.go
```diff
@@ -0,0 +1,159 @@
...
+	ExitRequested := false
+
+	ao := workflow.ActivityOptions{
+		StartToCloseTimeout: 10 * time.Second,
+	}
+	ctx = workflow.WithActivityOptions(ctx, ao)
+
+	// Listen to signals in a different goroutine
```
cretz, line 76: Why not the same goroutine with a selector shared by everything?

accumulator/accumulate_signals_workflow.go
```diff
@@ -0,0 +1,159 @@
...
+	workflow.Go(ctx, func(gCtx workflow.Context) {
+		for {
+			log.Debug("in workflow.go signals goroutine for{} \n")
```
cretz, line 79: Don't need to add newlines to log statements

cretz, line 79: You are inconsistent with whether you capitalize logs

accumulator/accumulate_signals_workflow.go
```diff
@@ -0,0 +1,159 @@
...
+	return totalLeft, nil
+}
+
+// AccumulateSignalsWorkflow workflow definition
+func AccumulateSignalsWorkflow(ctx workflow.Context, bucketKey string, greetingsCAN []AccumulateGreeting, uniqueGreetingKeysMapCAN map[string]bool) (allGreetings string, err error) {
+	log := workflow.GetLogger(ctx)
+	var a AccumulateGreeting
+	greetings := []AccumulateGreeting{}
+	greetings = append(greetings, greetingsCAN...)
+	unprocessedGreetings := []AccumulateGreeting{}
```
cretz, line 63:
```suggestion
	var unprocessedGreetings []AccumulateGreeting
```

No need to create it

accumulator/accumulate_signals_workflow.go
```diff
@@ -0,0 +1,159 @@
...
+		gotSignalBeforeTimeout, _ := workflow.AwaitWithTimeout(ctx, timeout, func() bool {
+			return len(unprocessedGreetings) > 0 || ExitRequested
+		})
+
+		// timeout happened without a signal coming in, so let's process the greetings and wrap it up!
+		if len(unprocessedGreetings) == 0 {
+			log.Debug("Into final processing, signal received? " + strconv.FormatBool(gotSignalBeforeTimeout)  + ", exit requested: " + strconv.FormatBool(ExitRequested) +"\n")
+			// do final processing
+			allGreetings = ""
+			err := workflow.ExecuteActivity(ctx, ComposeGreeting, greetings).Get(ctx, &allGreetings)
```
cretz, line 125: What if a signal comes in while processing this?

accumulator/accumulate_signals_workflow.go
```diff
@@ -0,0 +1,159 @@
...
+		/* process latest signal
+		 * Here is where we can process individual signals as they come in.
+		 * It's ok to call activities here.
+		 * This also validates an individual greeting:
+		 * - check for duplicates
+		 * - check for correct bucket 
+		 */
+		for len(unprocessedGreetings) > 0 {
+			ug := unprocessedGreetings[0]
+			unprocessedGreetings = unprocessedGreetings[1:]
```
cretz, line 143: The way Go works, this is actually a memory leak, but not a big deal because it's in a workflow. Regardless, I would just:

```go
toProcess := unprocessedGreetings
unprocessedGreetings = []AccumulateGreeting{}
for _, greeting := range toProcess {
```

accumulator/starter/main.go
```diff
@@ -0,0 +1,183 @@
...
+		nameIndex := rand.Intn(len(names))
+		
+		greeting := accumulator.AccumulateGreeting{
+			GreetingText: names[nameIndex],
+			Bucket: bucket,
+			GreetingKey: "key-" + fmt.Sprint(i),
+		}
+		time.Sleep(20 * time.Millisecond)
+
+		workflowId := WorkflowIDPrefix + "-" + bucket
```
cretz, line 63:
```suggestion
		workflowID := WorkflowIDPrefix + "-" + bucket
```

accumulator/starter/main.go
```diff
@@ -0,0 +1,183 @@
...
+    testSignalAfterExitSignal := !testSignalAfterWorkflowExit;
+    testDuplicate := true;
+    testIgnoreBadBucket := true;
+
+	// setup to send signals
+    bucket := "blue";
+    buckets := []string{"red", "blue", "green", "yellow"}
+    names := []string{"Genghis Khan", "Missy", "Bill", "Ted", "Rufus", "Abe"}
+
+	max_signals := 20
```
cretz, line 46: Go does not use underscores in variable names

accumulator/starter/main.go
```diff
@@ -0,0 +1,183 @@
...
+	})
+	if err != nil {
+		log.Fatalln("Unable to create client", err)
+	}
+	defer c.Close()
+
+	// setup which tests to run
+    // by default it will run an accumulation with a few (20) signals
+    // to a set of 4 buckets with Signal To Start
+    triggerContinueAsNewWarning := false;
```
cretz, line 32: Go does not use semicolons (usually)

accumulator/starter/main.go
```diff
@@ -0,0 +1,183 @@
...
+    testIgnoreBadBucket := true;
+
+	// setup to send signals
+    bucket := "blue";
+    buckets := []string{"red", "blue", "green", "yellow"}
+    names := []string{"Genghis Khan", "Missy", "Bill", "Ted", "Rufus", "Abe"}
+
+	max_signals := 20
+	if triggerContinueAsNewWarning {
+		max_signals = 10000
```
cretz, line 48: I think you can send less than half of this and get the same behavior

cretz (commented):
There's still a lot of little stuff, but I did a first skim for obvious things

accumulator/accumulate_signals_activities.go
joshmsmith: done thanks

joshmsmith: nice thanks

accumulator/accumulate_signals_activities.go
joshmsmith: good, thanks, fixed

accumulator/accumulate_signals_activities.go
joshmsmith: nice, thank you.

accumulator/starter/main.go
joshmsmith: changed, thanks

accumulator/starter/main.go
joshmsmith: ah, oops, copy/pasted/updated badly. fixed.

accumulator/accumulate_signals_workflow.go
joshmsmith: nice good changed, thanks.

accumulator/accumulate_signals_workflow.go
joshmsmith: nice, done, thank you.

accumulator/accumulate_signals_workflow.go
joshmsmith: changed, thank you.

accumulator/accumulate_signals_workflow.go
joshmsmith: done, this was a nice improvement, thanks.

accumulator/accumulate_signals_workflow.go
joshmsmith: changed, thank you.

accumulator/accumulate_signals_workflow.go
joshmsmith: changed all of them, thank you.

accumulator/starter/main.go
joshmsmith: good call, updated

accumulator/accumulate_signals_workflow.go
joshmsmith: done, this is great, thank you.

accumulator/accumulate_signals_workflow.go
joshmsmith: changed, thank you.

accumulator/accumulate_signals_workflow.go
joshmsmith: nice, fixed, thanks.

accumulator/accumulate_signals_workflow.go
joshmsmith: I wanted to demonstrate what happens in these signal-at-end-of-workflow cases. There's a test, Test_Signal_After_Exit(), and some code in the starter using a flag testSignalAfterExitSignal that works with this. In Go these signals are ignored, in Java the workflow goes back and processes them. 
Would it be good to add another check to len(unprocessedGreetings) == 0 before returning? I can comment why the sample does this, if you think it's worth doing so.

cretz: > In Go these signals are ignored

This is data loss which is usually not what is wanted with workflows like this

accumulator/accumulate_signals_workflow.go
joshmsmith: I was copying await_signals_workflow.go. I'm open to other approaches.

cretz: But that doesn't use a seaprate goroutine. There is probably no need for a separate goroutine here. Can just have a looping selector with `AddReceive` for the signals, `AddFuture` + `NewTimer` for the timeout, and then check `unprocessedGreetings` and `exitRequested` after the selector returns.

accumulator/accumulate_signals_workflow.go
```diff
@@ -0,0 +1,167 @@
...
+			log.Warn("Error awaiting signal")
+			return "", err
+		}
+		if timeout <= 0 {
+			// do final processing? or just check for signal one more time
+			log.Debug("No time left for signals, checking one more time")
+		}
+
+		log.Debug("Awaiting for " + timeout.String())
+		gotSignalBeforeTimeout, _ := workflow.AwaitWithTimeout(ctx, timeout, func() bool {
```
cretz, line 124: May be worth noting that you're adding two events for every task that has a signal here. You are adding a timer start and potentially a timer cancel. This is technically fine, but may be worth noting. Some users choose not to create a timer every time.

cretz:
@Quinn-With-Two-Ns - unsure if you want to give this a review

Quinn-With-Two-Ns:
@cretz I saw this, I just had no additional comments to add after yours

accumulator/accumulate_signals_workflow.go
joshmsmith: Thanks, changed.

accumulator/accumulate_signals_workflow.go
joshmsmith: updated to fix this, changed the relevant test.

accumulator/accumulate_signals_workflow.go
joshmsmith: I reworked this due to your feedback above about using a timer with a selector.

accumulator/accumulate_signals_workflow.go
```diff
@@ -0,0 +1,168 @@
+package accumulator
+
+import (
+	"strconv"
+	"time"
+
+	"go.temporal.io/sdk/workflow"
+)
+
+/**
```
cretz, line 10: Probably don't need to duplicate readme content in the file

accumulator/accumulate_signals_workflow.go
```diff
@@ -0,0 +1,168 @@
...
+	}
+	if startTime.IsZero() {
+		startTime = workflow.GetInfo(ctx).WorkflowStartTime // if you want to start from the time of the first signal, customize this
+	}
+	total := workflow.Now(ctx).Sub(startTime)
+	totalLeft := fromStartTimeout - total
+	if totalLeft <= 0 {
+		return 0, nil
+	}
+	if signalToSignalTimeout > totalLeft {
```
cretz, line 67: Don't you mean:

```suggestion
	if totalLeft > signalToSignalTimeout {
```

Since `signalToSignalTimeout` says it's the "maximum"?

accumulator/accumulate_signals_workflow.go
```diff
@@ -0,0 +1,168 @@
...
+ *   process individual events in the processGreeting() method.
+ *
+ * Because Temporal Workflows cannot have an unlimited size, Continue As New is used
+ *   to process more signals that may come in.
+ * You could create as many groupings as desired, as Temporal Workflows scale out to many workflows without limit.
+ * You could vary the time that the workflow waits for other signals, say for a day, or a variable time from first
+ *   signal with the GetNextTimeout() function.
+ */
+
+// signalToSignalTimeout is them maximum time between signals
```
cretz, line 29:
```suggestion
// signalToSignalTimeout is the maximum time between signals
```

accumulator/accumulate_signals_workflow.go
```diff
@@ -0,0 +1,168 @@
...
+	GreetingsList      []AccumulateGreeting
+	UniqueGreetingKeys map[string]bool
+	startTime          time.Time
+}
+
+// GetNextTimeout returns the maximum time for a workflow to wait for the next signal.
+// This waits for the greater of the remaining fromStartTimeout and  signalToSignalTimeout
+// fromStartTimeout and signalToSignalTimeout can be adjusted to wait for the right amount of time as desired
+// This resets with Continue As New
+func (a *AccumulateGreeting) GetNextTimeout(ctx workflow.Context, startTime time.Time, exitRequested bool) (time.Duration, error) {
```
cretz, line 55: Why is this a method on `AccumulateGreeting` struct? And why exported?

accumulator/accumulate_signals_workflow.go
cretz: Unresolving, this was not changed

accumulator/accumulate_signals_workflow.go
```diff
@@ -0,0 +1,168 @@
...
+			a = AccumulateGreeting{}
+		})
+		selector.AddReceive(workflow.GetSignalChannel(ctx, "exit"), func(c workflow.ReceiveChannel, more bool) {
+			c.Receive(ctx, nil)
+			exitRequested = true
+			cancelHandler() // cancel timer future
+			log.Debug("Exit Signal Received, more?: " + strconv.FormatBool(more) + "\n")
+		})
+
+		timerFuture := workflow.NewTimer(childCtx, timeout)
```
cretz, line 119: You should drain the signals before waiting. What happens if you get 100 signals at once? You are starting 100 timers erroneously. You should only wait when there are no signals in the channel.

accumulator/accumulate_signals_workflow.go
```diff
@@ -0,0 +1,168 @@
...
+		selector.AddFuture(timerFuture, func(f workflow.Future) {
+			log.Debug("Timer fired \n")
+		})
+
+		selector.Select(ctx)
```
cretz, line 124: Can you help me understand the goal here? Are you wanting to wait for the first signal or timer fired? Or do you want to collect all signals until timer reached and then process them? You need to loop on the select until timer fired if you want the latter and not cancel the timer on each signal.

accumulator/accumulate_signals_workflow.go
```diff
@@ -0,0 +1,168 @@
...
+		selector.AddReceive(workflow.GetSignalChannel(ctx, "exit"), func(c workflow.ReceiveChannel, more bool) {
+			c.Receive(ctx, nil)
+			exitRequested = true
+			cancelHandler() // cancel timer future
+			log.Debug("Exit Signal Received, more?: " + strconv.FormatBool(more) + "\n")
+		})
+
+		timerFuture := workflow.NewTimer(childCtx, timeout)
+		selector.AddFuture(timerFuture, func(f workflow.Future) {
+			log.Debug("Timer fired \n")
```
cretz, line 121: This can be a workflow cancel error, it should not be swallowed

accumulator/accumulate_signals_workflow.go
```diff
@@ -0,0 +1,168 @@
...
+		/* process latest signals
+		 * Here is where we can process individual signals as they come in.
+		 * It's ok to call activities here.
+		 * This also validates an individual greeting:
+		 * - check for duplicates
+		 * - check for correct bucket
+		 * Using update validation could improve this in the future
+		 */
+		toProcess := unprocessedGreetings
+		unprocessedGreetings = []AccumulateGreeting{}
```
cretz, line 151:
```suggestion
		unprocessedGreetings = nil
```

accumulator/accumulate_signals_workflow.go
```diff
@@ -0,0 +1,168 @@
...
+
+		if err != nil {
+			log.Error("Error calculating timeout")
+			return "", err
+		}
+		log.Debug("Awaiting for " + timeout.String())
+		selector.AddReceive(workflow.GetSignalChannel(ctx, "greeting"), func(c workflow.ReceiveChannel, more bool) {
+			c.Receive(ctx, &a)
+			unprocessedGreetings = append(unprocessedGreetings, a)
+			log.Debug("Signal Received with text: " + a.GreetingText + ", more?: " + strconv.FormatBool(more) + "\n")
```
cretz, line 108:
```suggestion
			log.Debug("Signal Received", "Text", a.GreetingText, "More", more)
```

Loggers are expected to take key-value state. Also, don't append newlines to log statements.

accumulator/accumulate_signals_workflow.go
```diff
@@ -0,0 +1,168 @@
...
+func AccumulateSignalsWorkflow(ctx workflow.Context, greetings GreetingsInfo) (allGreetings string, err error) {
+	log := workflow.GetLogger(ctx)
+	var a AccumulateGreeting
+	if greetings.GreetingsList == nil {
+		greetings.GreetingsList = []AccumulateGreeting{}
+	}
+	if greetings.UniqueGreetingKeys == nil {
+		greetings.UniqueGreetingKeys = make(map[string]bool)
+	}
+	var unprocessedGreetings []AccumulateGreeting
```
cretz, line 83: The way the code is currently written, this will only ever have one value in it, so why make it a slice?

accumulator/accumulate_signals_workflow.go
```diff
@@ -0,0 +1,168 @@
...
+	return totalLeft, nil
+}
+
+// AccumulateSignalsWorkflow workflow definition
+func AccumulateSignalsWorkflow(ctx workflow.Context, greetings GreetingsInfo) (allGreetings string, err error) {
+	log := workflow.GetLogger(ctx)
+	var a AccumulateGreeting
+	if greetings.GreetingsList == nil {
+		greetings.GreetingsList = []AccumulateGreeting{}
+	}
```
cretz, line 79:
```suggestion
```

This is not needed

accumulator/starter/main.go
```diff
@@ -0,0 +1,201 @@
...
+		if err != nil {
+			log.Fatalln("Unable to signal workflow "+we.GetID(), err)
+		}
+		log.Println(we.GetID() + ": Sent exit")
+		// Signals after this test sending more signals after workflow exit
+		time.Sleep(5 * time.Millisecond)
+	}
+
+	
+
```
cretz, line 157: Go format usually fixes all of these extra lines, is it run?

accumulator/starter/main.go
```diff
@@ -0,0 +1,201 @@
+package main
+
+import (
+	"context"
+	"fmt"
+	"log"
+	"time"
+
```
cretz, line 8: Usually stdlib imports are grouped together and modern autoformat fixes this usually

accumulator/starter/main.go
```diff
@@ -0,0 +1,201 @@
...
+	"math/rand"
+
+	accumulator "github.com/temporalio/samples-go/accumulator"
```
cretz, line 11:
```suggestion
	"github.com/temporalio/samples-go/accumulator"
```

No need to alias

accumulator/starter/main.go
```diff
@@ -0,0 +1,201 @@
...
+	// The client is a heavyweight object that should be created once per process.
+	c, err := client.Dial(client.Options{
+		HostPort: client.DefaultHostPort,
+	})
+	if err != nil {
+		log.Fatalln("Unable to create client", err)
+	}
+	defer c.Close()
+
+	// setup which tests to run
```
cretz, line 29: You are inconsistent with whether you capitalize/punctuate sentences. Also, this is a bit confusing to follow.

joshmsmith:
closing. with the other pending improvements (message passing) and update coming, decided this was no longer needed.

PR #359, opened 2024-07-24
Add Nexus sample

bergundy:


Quinn-With-Two-Ns:
Can we include the sample in the https://github.com/temporalio/samples-go/blob/main/README.md

nexus/README.md
```diff
@@ -0,0 +1,96 @@
+# nexus-hello-api
```
Quinn-With-Two-Ns, line 1: Can we include a overview of what nexus is or link out to something that explains what this nexus is?

nexus/caller/app/workflows.go
```diff
@@ -0,0 +1,43 @@
...
+	"github.com/temporalio/samples-go/nexus/service"
+	"go.temporal.io/sdk/workflow"
+)
+
+const (
+	TaskQueue    = "my-caller-workflow-task-queue"
+	endpointName = "my_nexus_endpoint_name"
+)
+
+func EchoCallerWorkflow(ctx workflow.Context, message string) (string, error) {
```
Quinn-With-Two-Ns, line 13: Other samples just put the workflow in the samples root folder, don't think we need to do anything different here.

https://github.com/temporalio/samples-go/blob/main/helloworld/helloworld.go

nexus/README.md
Quinn-With-Two-Ns: And explain what the sample is showing

Quinn-With-Two-Ns:
Can we also show testing? Generally our samples how how to use and test a feature.

nexus/README.md
```diff
@@ -0,0 +1,96 @@
...
+
+```
+./bin/temporal server start-dev --http-port 7243 --dynamic-config-value system.enableNexus=true
+```
+
+### Initialize environment
+
+In a separate terminal window
+
+#### Create caller and target namespaces
```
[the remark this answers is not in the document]
bergundy: No lag, there's a namespace read-through cache.

nexus/caller/app/workflows.go
```diff
@@ -0,0 +1,43 @@
...
+	c := workflow.NewNexusClient(endpointName, service.HelloServiceName)
+
+	fut := c.ExecuteOperation(ctx, service.EchoOperationName, service.EchoInput{Message: message}, workflow.NexusOperationOptions{})
```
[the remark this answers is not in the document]
bergundy: Perhaps but there's not that much value if we don't provide type safe APIs on top of that.

nexus/service/description.md
```diff
@@ -0,0 +1,8 @@
+Service Name:
```
Quinn-With-Two-Ns, line 1: Is there any documentation we can link users to about this file format? I imagine a user seeing nexus for the first time might be confused on what they can do here.

nexus/caller/app/workflows.go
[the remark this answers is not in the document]
cretz: Similar could be said for activities and child workflows and such. It is about what we want to encourage. Also arguably you could check argument types on the way out.

nexus/service/description.md
bergundy: There's no public documentation yet. We plan to have markdown support in the UI and a template that will direct users what to put here.

cretz: A template? Is this not freeform markdown that can have anything? Will there be an expectation of some format/values _inside_ the markdown?

bergundy: It is free form but we'd like to give some guidance on what would be useful to put in it.
This is likely a temporary step until we have code generation and a way to publish schemas to be explored in the UI.

cretz: Makes sense. In my mind whatever is useful to them is useful to them and we don't need to mention what is useful to them, but I guess it doesn't matter much

nexus/README.md
```diff
@@ -0,0 +1,112 @@
...
+- [caller](./caller) - caller workflows, worker, and starter
+- [handler](./handler) - handler workflow, operations, and worker
+- [options](./options) - command line argument parsing utility
+
+## Getting started locally
+
+### Get `temporal` CLI `v0.14.0-nexus.0` to enable local development
+
+Go to the [CLI release page](https://github.com/temporalio/cli/releases/tag/v0.14.0-nexus.0) and download
+an artifact for your OS and architecture.
```
cretz, line 27: May also need to say it needs to be extracted. Also samples below assume it's in `./bin` which is confusing (and not very cross-platform). But all of this will go away when it's in stable server anyways, so no biggie (we need to set a reminder to come back and update this).

cretz (approved):

Quinn-With-Two-Ns (approved):

PR #362, opened 2024-08-22
remove stale import from helloworld

yuandrew:
## What was changed

The import doesn't seem to be necessary, and the TODO is 2 years old. Removing

## Why?

Clean things up

## Checklist

1. Closes 

2. How was this tested:

`make ci-build` passes locally

3. Any docs updates needed?


cretz (approved):
LGTM. Yeah now that this is tagged, the `go install` for `@latest` in the Makefile works without this. Merge at will.

PR #364, opened 2024-09-24
Safe message passing sample

Quinn-With-Two-Ns:
Add Safe message passing sample based on https://github.com/temporalio/samples-python/tree/main/message_passing/safe_message_handlers


safe_message_handler/README.md
cretz, line 1: Should we add a top-level README link to this sample?

cretz (approved):
Did not review very deeply, but LGTM on the surface

PR #365, opened 2024-10-01
adding snipsync to nexus go samples 

jsundai:
## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/nexus/caller/starter/main.go
+++ b/nexus/caller/starter/main.go
@@ -1,4 +1,4 @@
-// @@@SNIPSTART startercaller-nexus
+// @@@SNIPSTART samples-go-nexus-caller-starter
 package main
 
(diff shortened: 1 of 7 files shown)


nexus/caller/starter/main.go
```diff
@@ -1,3 +1,4 @@
+// @@@SNIPSTART startercaller-nexus
```
bergundy, line 1: You'll want to prefix all of these snippets with `samples-go-` as done for the rest of the snippets in this repo as snipsync doesn't use the repo name to prefix the snippet name.

jsundai: Sounds good.

bergundy (approved):

PR #366, opened 2024-10-02
Early-return sample

stephanos:
## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


early-return/workflow.go
```diff
@@ -0,0 +1,77 @@
...
+	var initErr error
+	var initDone bool
+	logger := workflow.GetLogger(ctx)
+
+	if err := workflow.SetUpdateHandler(
+		ctx,
+		UpdateName,
+		func(ctx workflow.Context) error {
+			condition := func() bool { return initDone }
+			if completed, err := workflow.AwaitWithTimeout(ctx, updateTimeout, condition); err != nil {
```
drewhoskins-temporal, line 38: nit: since update is more of an implementation detail here, suggest something like `earlyReturnTimeout`.

early-return/workflow.go
stephanos, line 1: To be honest, I feel very unsure about the various timeouts here and on the caller side. I'm not quite sure how to think about.

early-return/workflow.go
```diff
@@ -0,0 +1,77 @@
...
+)
+
+// Workflow processes a transaction in two phases. First, the transaction is initialized, and if successful,
+// it proceeds to completion. However, if initialization fails - due to validation errors or transient
+// issues (e.g., network connectivity problems) - the transaction is cancelled.
+//
+// By utilizing Update-with-Start, the client can initiate the workflow and immediately receive the result of
+// the initialization in a single round trip, even before the transaction processing completes. The remainder
+// of the transaction is then processed asynchronously.
+func Workflow(ctx workflow.Context, transactionId, fromAccount, toAccount string, amount float64) error {
```
Quinn-With-Two-Ns, line 28: For workflows with multiple input parameters we generally tell users to always use structs over multiple parameters to make it easier to update the workflow in the future

early-return/starter/main.go
```diff
@@ -0,0 +1,56 @@
...
+		log.Fatalln("Error obtaining update handle:", err)
+	}
+
+	err = updateHandle.Get(ctxWithTimeout, nil)
+	if err != nil {
+		// NOTE: If the error is retryable, a retry attempt must use a unique workflow ID.
+		log.Fatalln("Error obtaining update result:", err)
+	}
+
+	log.Println("Transaction completed successfully")
```
Quinn-With-Two-Ns, line 53: nit: this log is a bit misleading since the transaction could still be running right?

early-return/activity.go
```diff
@@ -0,0 +1,33 @@
...
+		return errors.New("invalid toAccount")
+	}
+	if amount == 0 {
+		return errors.New("invalid amount")
+	}
+	logger.Info("Transaction initialized")
+	return nil
+}
+
+func CancelTransaction(ctx context.Context, transactionId string) {
```
Quinn-With-Two-Ns, line 25: nit: could add some delay here and to `CompleteTransaction` so "early return" is returning an observable amount before the real workflow

early-return/workflow.go
stephanos: 👍 Makes a lot of sense. Done.

early-return/starter/main.go
stephanos: True, I've rephrased it now.

early-return/activity.go
stephanos: 👍 Good idea

early-return/workflow.go
```diff
@@ -0,0 +1,82 @@
...
+	var initErr error
+	var initDone bool
+	logger := workflow.GetLogger(ctx)
+
+	if err := workflow.SetUpdateHandler(
+		ctx,
+		UpdateName,
+		func(ctx workflow.Context) error {
+			condition := func() bool { return initDone }
+			if completed, err := workflow.AwaitWithTimeout(ctx, earlyReturnTimeout, condition); err != nil {
```
cretz, line 45: I think you should leave how long a caller is willing to wait for the initial update up to them unless it's really important to differentiate start-to-update timeout from schedule-to-update timeout.

early-return/workflow.go
```diff
@@ -0,0 +1,82 @@
...
+	"go.temporal.io/sdk/workflow"
+)
+
+const (
+	UpdateName    = "early-return"
+	TaskQueueName = "early-return-tq"
+)
+
+var (
+	activityTimeout    = 2 * time.Second
```
cretz, line 17: I understand we want to demonstrate low latency, but this is a pretty aggressive timeout. But it is probably ok.

early-return/workflow.go
```diff
@@ -0,0 +1,82 @@
...
+
+	"go.temporal.io/sdk/workflow"
+)
+
+const (
+	UpdateName    = "early-return"
+	TaskQueueName = "early-return-tq"
+)
+
+var (
```
cretz, line 16: These can be constants too btw

early-return/workflow.go
```diff
@@ -0,0 +1,82 @@
...
+	activityTimeout    = 2 * time.Second
+	earlyReturnTimeout = 5 * time.Second
+)
+
+type Transaction struct {
+	ID          string
+	FromAccount string
+	ToAccount   string
+	Amount      float64
```
cretz, line 25: Usually a sin these days when talking about money to use floats

early-return/workflow.go
```diff
@@ -0,0 +1,82 @@
...
+
+	activityOptions := workflow.WithLocalActivityOptions(ctx, workflow.LocalActivityOptions{
+		ScheduleToCloseTimeout: activityTimeout,
+	})
+	initErr = workflow.ExecuteLocalActivity(activityOptions, InitTransaction, tx).Get(ctx, nil)
+	initDone = true
+
+	// Phase 2: Complete or cancel the transaction asychronously.
+	activityCtx := workflow.WithActivityOptions(ctx, workflow.ActivityOptions{
+		StartToCloseTimeout: 10 * time.Second,
```
cretz, line 69: It is confusing that the `activityTimeout` global is only used once and is for local activity timeout and this isn't even a global and is actually an activity timeout. Arguably there is no need for these single-use globals, but if there is, consider consistently using globals for these.

early-return/workflow.go
```diff
@@ -0,0 +1,82 @@
...
+	); err != nil {
+		return err
+	}
+
+	// Phase 1: Initialize the transaction synchronously.
+	//
+	// By using a local activity, an additional server roundtrip is avoided.
+	// See https://docs.temporal.io/activities#local-activity for more details.
+
+	activityOptions := workflow.WithLocalActivityOptions(ctx, workflow.LocalActivityOptions{
```
cretz, line 61: I doubt you'll want the default retry options with such an aggressive schedule to close of 2s

cretz (commented):
Would recommend tests when this becomes a non-draft PR

early-return/workflow.go
stephanos: Yeah, the idea for the sample is to push the low latency story; and this is actually on the upper end of what I'm hearing for customer use cases. 100-500ms for the entire Update is what's often needed.

early-return/workflow.go
stephanos: I had the same thought at first, but I figured it's more intuitive this way in the context of a sample? If this has shifted and I didn't get the memo, I'm happy to change it. I didn't want to add extra complexity/confusion.

early-return/workflow.go
```diff
@@ -0,0 +1,82 @@
...
+			} else if !completed {
+				return errors.New("update timed out")
+			}
+			return initErr
+		},
+	); err != nil {
+		return err
+	}
+
+	// Phase 1: Initialize the transaction synchronously.
```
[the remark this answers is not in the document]
stephanos: In all my examples until now, I've actually had it flipped. Drew convinced me to do it the other way around, but I'm not quite sure anymore why.

early-return/workflow.go
stephanos: Right; I've just made them global to make it easier to see at a glance what the timeouts are without reading line-by-line.

early-return/workflow.go
```diff
@@ -0,0 +1,82 @@
...
+
+	// Phase 2: Complete or cancel the transaction asychronously.
+	activityCtx := workflow.WithActivityOptions(ctx, workflow.ActivityOptions{
+		StartToCloseTimeout: 10 * time.Second,
+	})
+	if initErr != nil {
+		logger.Info("cancelling transaction due to error: %v", initErr)
+
+		// Transaction failed to be initialized or not quickly enough; cancel the transaction.
+		return workflow.ExecuteActivity(activityCtx, CancelTransaction, tx).Get(ctx, nil)
```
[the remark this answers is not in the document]
stephanos: I see your point. My only worry is that users wouldn't be able to distinguish between "failed to init" and "failed to cancel/complete" - which might require very different actions.

early-return/workflow.go
cretz: If that's the case, may not want a default retry policy with an initial interval of a second. May actually want max attempts as 1.

early-return/workflow.go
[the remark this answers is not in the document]
Quinn-With-Two-Ns: Since updates are not durable on admitted that is why we did it this way so this is the only safe way to write this type of workflow. if I remember correctly

early-return/workflow.go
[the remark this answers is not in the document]
cretz: What I mean here is that if `CancelTransaction` succeeds, operators have no observability into the failure because you have logged the failure as info and did not fail the workflow. Usually with compensating actions, you want to propagate the original failure, not log-and-swallow.

early-return/workflow.go
cretz: I think int would be better, I doubt it adds complexity. We do this in our tutorial too at https://github.com/temporalio/money-transfer-project-template-go/blob/2bb1672af07cb76d449f14beb046e412f44a7afb/shared.go#L12

stephanos: I see 👍 I'll change it. I thought this would go into the idea of representing cents, too, but the linked example just uses "250" without any denomination.

early-return/workflow.go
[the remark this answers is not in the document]
Quinn-With-Two-Ns: It isn't more preferable this is the only safe way to write it since update with start is not transactional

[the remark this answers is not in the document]
drewhoskins-temporal: My main reasoning is what chad said: you want the workflow to function properly when the client doesn't call it with an update.
Quinn's reasoning makes sense too.

early-return/workflow.go
drewhoskins-temporal: I would be as lax as you can on these timeouts.  Don't want to fail requests that would have otherwise succeeded.  So if the overall task timeout is 10s, maybe give it 9 or 10s?

early-return/workflow.go
drewhoskins-temporal: I would be as lax as you can on these timeouts. Don't want to unnecessarily fail requests that would have otherwise succeeded. So if the overall workflow task timeout is 10s, maybe give it 9 or 10s?

early-return/workflow.go
drewhoskins-temporal: nit: can you use a different, longer timeout here?  since this is the async part, and I think 10s was used elsewhere?  I think 30s is a fairly standard timeout for a rando activity.

early-return/workflow.go
```diff
@@ -0,0 +1,79 @@
...
+	"time"
+
+	"go.temporal.io/sdk/workflow"
+)
+
+const (
+	UpdateName         = "early-return"
+	TaskQueueName      = "early-return-tq"
+	activityTimeout    = 2 * time.Second
+	earlyReturnTimeout = 5 * time.Second
```
drewhoskins-temporal, line 15: This is aggressive.  Our usual advice is to have more generous timeouts but to monitor latencies and keep them low.
Think about sitting at a computer waiting for an operation to complete.  If it was going to take more than 5 seconds, would you want it to fail saying "maybe that worked" ?  Or would you rather wait longer?  You'd probably be willing to wait a little longer.
Many client-facing RPC servers time out after around 30 seconds, and so these sorts of timeouts can be calibrated to be shorter than that.  
Here's a draft of a comment: 

One common heuristic: Calibrate this number relative to your overall remaining client timeout.  So, if your client will timeout in 29 more seconds, you might choose 28s to give time to return and report the correct error.
In general, err on the generous side so as not to fail operations that would have succeeded.

drewhoskins-temporal: Curious if @cretz 's advice would be similar.

early-return/workflow.go
cretz: I'd be ok if you documented that it was in cents or added `Cents` to the field or something

early-return/workflow.go
cretz: Completely situational I think. No strong opinion. Arguably the caller should determine how long they're willing to wait. A timer inside a workflow does not account for, say, the workflow being slow to start.

early-return/workflow.go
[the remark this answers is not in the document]
cretz: > It isn't more preferable this is the only safe way to write it since update with start is not transactional

Do we at least guarantee the update and the start are in the same task? If we don't, all latency bets are off anyways. But whether primary workflow waits on update or update waits on primary workflow is immaterial I'd think.

early-return/workflow.go
```diff
@@ -0,0 +1,82 @@
...
+// of the transaction is then processed asynchronously.
+func Workflow(ctx workflow.Context, tx Transaction) error {
+	var initErr error
+	var initDone bool
+	logger := workflow.GetLogger(ctx)
+
+	if err := workflow.SetUpdateHandler(
+		ctx,
+		UpdateName,
+		func(ctx workflow.Context) error {
```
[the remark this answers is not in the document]
stephanos: Good idea 👍 Only part I don't quite follow is the "run as a method". Is that possible in Go? I saw an error when trying to register the workflow method on the struct and couldn't find any example doing that (only for activities).

early-return/workflow.go
stephanos: Okay; I thought there was a risk that the caller might accidentally wait indefinitely if they don't specify a deadline on their end. But I just turned off the worker, and the ExecuteWorkflow request times out after 10s, even though I'm using `context.Background()`. I wasn't aware that timeout existed.

early-return/workflow.go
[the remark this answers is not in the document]
stephanos: We do guarantee it for Update-with-Start.

early-return/workflow.go
```diff
@@ -0,0 +1,112 @@
...
+	if err := workflow.ExecuteActivity(activityCtx, tx.CompleteTransaction).Get(ctx, nil); err != nil {
+		return fmt.Errorf("completing the transaction failed: %w", err)
+	}
+
+	return nil
+}
+
+func (tx *Transaction) ReturnInitResult(ctx workflow.Context) error {
+	if err := workflow.Await(ctx, func() bool { return tx.initDone }); err != nil {
+		return fmt.Errorf("transaction init cancelled: %w", err)
```
stephanos, line 85: AFAICT, this is the only untested line of the workflow. I'm not quite sure how to use the Go SDK testing env to trigger this case.

early-return/workflow.go
stephanos: - I've removed the "Await" timeout (`earlyReturnTimeout`) since it should be cancelled when the context times out. (see [other convo](https://github.com/temporalio/samples-go/pull/366#discussion_r1786974554))

- I've bumped the local activity timeout to 10s now; and the async activity to 30s.

Quinn-With-Two-Ns:
We should link this sample from the readme as we do for all samples https://github.com/temporalio/samples-go/blob/main/README.md

early-return/workflow_test.go
```diff
@@ -0,0 +1,102 @@
...
+func Test_CompleteTransaction(t *testing.T) {
+	testSuite := &testsuite.WorkflowTestSuite{}
+	env := testSuite.NewTestWorkflowEnvironment()
+
+	tx := Transaction{ID: uuid.New(), SourceAccount: "Bob", TargetAccount: "Alice", Amount: 100}
+	env.RegisterActivity(tx.InitTransaction)
+	env.RegisterActivity(tx.CompleteTransaction)
+
+	uc := &updateCallback{}
+	env.RegisterDelayedCallback(func() {
```
Quinn-With-Two-Ns, line 22: nit: would add a comment explaining this will guarantee the update is sent in the first WFT.

early-return/workflow.go
Quinn-With-Two-Ns: You would need to test cancellation using https://pkg.go.dev/go.temporal.io/sdk@v1.29.1/internal#TestWorkflowEnvironment.CancelWorkflow no? Not saying you need to for this sample though.

Quinn-With-Two-Ns (approved):

early-return/workflow.go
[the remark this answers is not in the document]
cretz: :+1: Then yeah I think it's probably just semantics on which coroutine waits on the other and probably doesn't matter

early-return/workflow_test.go
drewhoskins-temporal: FWIW I'm confused even after Quinn's comment.  Why is this test seemingly not using the normal interface for UwS?
(I assume you have a good reason and that there's nothing fixable in this PR.)

drewhoskins-temporal: Is it not backed by the Java test service?

early-return/workflow.go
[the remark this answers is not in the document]
cretz: > Only part I don't quite follow is the "run as a method". Is that possible in the Go SDK?

You'd do the wrapping w/ a one-liner, so something like:

```go
type myWorkflow struct { SomeState }

func MyWorkflow(ctx workflow.Context, someState SomeState) (*SomeResult, error) {
    return myWorkflow{ someState }.run(ctx)
}

func (m *myWorkflow) run(ctx workflow.Context) (*SomeResult, error) {
    // This kind of setup could go into a newMyWorkflow(...) call instead of in here
    if err := workflow.SetUpdateHandler(ctx, "myUpdate", m.myUpdate); err != nil {
        return nil, err
    }
    panic("TODO")
}

func (m *myWorkflow) myUpdate(ctx workflow.Context, someParam SomeParam) (*SomeUpdateResult, error) {
    panic("TODO")
}
```

early-return/workflow.go
stephanos: Ah, that's the missing piece! Yeah, I agree, it's prob fine.

early-return/workflow_test.go
stephanos: Correct; Go SDK has it's own time-skipping test server, and it has its own APIs. We haven't added an UwS API since this approach here works, too.

early-return/workflow.go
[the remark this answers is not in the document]
stephanos: 👍 I thought I might have missed a trick to do it with a single method

early-return/workflow.go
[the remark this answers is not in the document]
dandavison: Another strong reason to do it this way is that, if the update did the init, then the workflow author has to make sure the workflow is correct in the face of multiple calls to the update handler, i.e. normal updates being sent subsequent to the update with start. But with all steps in the main workflow, multiple calls to the update handler are automatically correct.

PR #373, opened 2024-12-03
blob storage codec sample

kevinawoo:
## What was changed
Add a sample to show how to use the codec to store payloads between the Client, Server, and Workers.

## Why?
gRPC has a 4MB payload limit. We've seen this limit get hit constantly. This provides a sample of how to use a blob-store to pass large payloads if we exceed a certain size.

## Checklist
1. How was this tested:
- run the worker and starter to generate a execution
- open the UI and inspect the payloads. Half the payloads are plain text, the other half will be stored in the blob store
- run the codec-server, connect it to the UI. All payloads should be rendered as plain text

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,6 +82,10 @@ Each sample demonstrates one feature of the SDK, together with tests.
   server to decode payloads for display in tctl and Temporal Web. This setup can be used for any kind of codec, common
   examples are compression or encryption.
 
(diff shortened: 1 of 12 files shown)


kevinawoo:
closing as this is not recommended because it introduces a lot of network activity during reply. An alternative solution of explicitly calling upload/download is preferred. I'll work on this sample

PR #371, opened 2024-11-22
Switch to using TestUpdateCallback

yuandrew:
## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

PR #374, opened 2024-12-12
Bump SDK to 1.31

stephanos:
## What was changed

- [x] bumped SDK to new version
- [x] updated UwS API
- [x] extended early return example


reqrespupdate/workflow_test.go
stephanos, line 1: from https://github.com/temporalio/samples-go/pull/371

stephanos: Hm, I copied the code verbatim but it fails. would you know why, @yuandrew?

early-return/starter/main.go
stephanos, line 1: @dandavison I revamped this based on your notes and the Java sample.

reqrespupdate/workflow_test.go
yuandrew: hmm, looking, it's not immediately clear why this would suddenly fail, I tested this locally when making the Go SDK change, and it should be a pretty basic UX change, shouldn't be any change to functionality. 

Looking at the logs, only 1 activity is completing, instead of the 500 the test expects

reqrespupdate/workflow_test.go
```diff
@@ -38,19 +20,19 @@ func TestUppercaseWorkflow(t *testing.T) {
 	for i := 0; i < 550; i++ {
 		i := i
 		env.RegisterDelayedCallback(func() {
-			env.UpdateWorkflow(reqrespupdate.UpdateHandler, "test id", &updateCallback{
-				accept: func() {
+			env.UpdateWorkflow(reqrespupdate.UpdateHandler, "test id", &testsuite.TestUpdateCallback{
```
yuandrew, line 23:
```suggestion
			env.UpdateWorkflow(reqrespupdate.UpdateHandler, fmt.Sprintf("test id_%d", i), &testsuite.TestUpdateCallback{
```

reqrespupdate/workflow_test.go
yuandrew: Ok, found the issue, this was due to https://github.com/temporalio/sdk-go/pull/1695 changing test behavior to more closely match how server handles duplicate update IDs. Put the suggested fix in a separate comment.

Thanks for catching this!

stephanos: @yuandrew thank you so much!

yuandrew (approved):

early-return/starter/main.go
```diff
@@ -20,38 +21,55 @@ func main() {
...
+		// a server-side failure (e.g. failed to start workflow, or exceeded
+		// limit on concurrent update per workflow execution).
+		log.Fatalln("Error issuing update-with-start:", err)
+	}
+
+	log.Println("Started workflow",
+		"WorkflowID:", updateHandle.WorkflowID(),
+		"RunID:", updateHandle.RunID())
+
+	var txResult earlyreturn.Transaction
```
dandavison, line 56: I think that we should name this something like `earlyreturnResult`, or `transactionConfirmation`, or something else along those lines. The key point we want to get across is that the transaction hasn't completed yet, but `txResult` suggests otherwise.

stephanos: I see what you mean. I can rename it 👍 The [java-sample has the same issue](https://github.com/temporalio/samples-java/blob/main/core/src/main/java/io/temporal/samples/earlyreturn/EarlyReturnClient.java#L53C5-L53C13).

PR #379, opened 2025-01-07
New shopping cart sample

yuandrew:
## What was changed

Added a new shopping cart app to demonstrate using web sockets with temporal

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


shoppingcart/server/main.go
```diff
@@ -0,0 +1,137 @@
...
+	cartState      = make(map[string]int)
+	workflowClient client.Client
+	itemCosts      = map[string]int{
+		"apple":      2,
+		"banana":     1,
+		"watermelon": 5,
+		"television": 1000,
+		"house":      10000000,
+		"car":        50000,
+		"binder":     10,
```
dandavison, line 24: I wonder whether these should be cents, in the interests of making the sample more like real code.

early-return/README.md
```diff
@@ -5,9 +5,16 @@ This sample demonstrates an early-return from a workflow.
...
+for Update-with-Start being used for lazy initialization.
+
 ### Steps to run this sample:
 1) Run a [Temporal service](https://github.com/temporalio/samples-go/tree/main/#how-to-use).
-2) Run the following command to start the worker
+
+   NOTE: frontend.enableExecuteMultiOperation=true must be configured for the server
+in order to use Update-with-Start. 
+
+5) Run the following command to start the worker
```
Quinn-With-Two-Ns, line 17: What happened to step 2-4?

Quinn-With-Two-Ns: Oh maybe github is wrong here because the rich diff shows this as 2)

shoppingcart/workflow.go
```diff
@@ -0,0 +1,45 @@
...
+	TaskQueueName = "shopping-cart-tq"
+)
+
+type CartSignalPayload struct {
+	Action   string `json:"action"` // "add" or "remove"
+	ItemID   string `json:"item_id"`
+	Quantity int    `json:"quantity"`
+}
+
+type CartState map[string]int // itemID -> quantity
```
Quinn-With-Two-Ns, line 18: nit: I would make `CartState` a struct with a map inside. I suggested this because it would be more externsible in the future if we wanted to add additional data to the cart state.

shoppingcart/workflow.go
```diff
@@ -0,0 +1,45 @@
...
+		if itemID != "" {
+			if actionType == "add" {
+				cart[itemID] += 1
+			} else if actionType == "remove" {
+				cart[itemID] -= 1
+				if cart[itemID] <= 0 {
+					delete(cart, itemID)
+				}
+			} else {
+				logger.Error("Unknown action type:", actionType)
```
Quinn-With-Two-Ns, line 35: could use the validator to reject unknown action types

shoppingcart/workflow.go
```diff
@@ -0,0 +1,45 @@
...
+			}
+		}
+		return cart, nil
+	}); err != nil {
+		return err
+	}
+
+	// Keep workflow alive to continue to listen receive updates.
+	return workflow.Await(ctx, func() bool { return false })
```
Quinn-With-Two-Ns, line 44: hm, this would mean the workflow would never complete or CAN, I don't think that is correct because workflows have a history size limit and a total. I think we should add CAN support here and maybe some way to "complete" the workflow. If this is a shopping cart and it is checked out the cart is no longer needed.

yuandrew: What's the best practice for waiting for updates for CAN? Is there something better than `workflow.Sleep` I should use here?
```
for {
	if workflow.GetInfo(ctx).GetContinueAsNewSuggested() {
		logger.Info("Continuing as new")
		return workflow.NewContinueAsNewError(ctx, CartWorkflow)
	}
	if checkout {
		return nil
	}

	err := workflow.Sleep(ctx, time.Second)
	if err != nil {
		return err
	}
}
```

yuandrew: actually, I think i answered my own question, 
`err := workflow.Await(ctx, func() bool { return workflow.GetInfo(ctx).GetContinueAsNewSuggested() || checkout })` seems like this would work better

Quinn-With-Two-Ns: Yes, but keep in mind you don't want to CAN if you have any inflight updateshttps://docs.temporal.io/encyclopedia/workflow-message-passing#finishing-message-handlers

early-return/README.md
yuandrew: good catch! it was misnumbered, but it renders at 2 🤷‍♂️

shoppingcart/server/main.go
```diff
@@ -0,0 +1,151 @@
...
+	if err != nil {
+		// For example, a client-side validation error (e.g. missing conflict
+		// policy or invalid workflow argument types in the start operation), or
+		// a server-side failure (e.g. failed to start workflow, or exceeded
+		// limit on concurrent update per workflow execution).
+		log.Fatalln("Error issuing update-with-start:", err)
+	}
+
+	// If someone has checked out their cart, this completes the workflow.
+	// We then want to create a new workflow for the next user to shop.
```
dandavison, line 134: I'm not qute following this -- it makes it sound like the site would only handle one concurrent shopper. In my experience with e-commerce sites, a "session ID" is obtained from a cookie submitted by the end user client, and there's a database table containing session info keyed by session ID, from which an active shopping cart can be obtained. So in the Python sample I faked this by hard-coding a session ID, and using it as the workflow ID for the UwS operations. https://github.com/temporalio/samples-python/blob/main/message_passing/update_with_start/lazy_initialization/starter.py#L34

shoppingcart/workflow.go
```diff
@@ -0,0 +1,71 @@
...
+var (
+	UpdateName    = "shopping-cart"
+	TaskQueueName = "shopping-cart-tq"
+)
+
+type CartSignalPayload struct {
+	Action   string `json:"action"` // "add" or "remove"
+	ItemID   string `json:"item_id"`
+	Quantity int    `json:"quantity"`
+}
```
dandavison, line 17: unused

shoppingcart/server/main.go
```diff
@@ -0,0 +1,151 @@
...
+	if actionType == "checkout" {
```
dandavison, line 135: I'm thinking that `checkout` should be a distinct operation from `addItem`. checkout doesn't need to be an updateWithStart: it could be an update or a signal.

shoppingcart/server/main.go
```diff
@@ -0,0 +1,151 @@
...
+	"github.com/temporalio/samples-go/shoppingcart"
+	enumspb "go.temporal.io/api/enums/v1"
+	"go.temporal.io/sdk/client"
+	"log"
+	"net/http"
+	"sort"
+)
+
+var (
+	cartState      = shoppingcart.CartState{Items: make(map[string]int)}
```
dandavison, line 16: I don't think cart state should be held in a global variable accessible in the client web app. The point of modeling a cart as a temporal workflow is that the storage, and point-of-truth, for the cart contents is the workflow itself. Using this global variable makes that less clear to the readers IMO. But I'm just proposing a small change: have `listHandler` take the cart state as an argument.

shoppingcart/server/main.go
```diff
@@ -0,0 +1,151 @@
...
+	id := r.URL.Query().Get("id")
+
+	updateWithStartCart(actionType, id)
+
+	if actionType != "" {
+		listHandler(w, r)
+	}
+}
+
+func updateWithStartCart(actionType string, id string) {
```
dandavison, line 104: See comment above: IMO this function should return `CartState`

shoppingcart/workflow.go
```diff
@@ -0,0 +1,71 @@
...
+	cart := CartState{make(map[string]int)}
+	logger := workflow.GetLogger(ctx)
+	var checkout bool
+
+	if err := workflow.SetUpdateHandlerWithOptions(ctx, UpdateName, func(ctx workflow.Context, actionType string, itemID string) (CartState, error) {
+		logger.Info("Received update,", actionType, itemID)
+		if actionType == "checkout" {
+			checkout = true
+		}
+		if itemID != "" {
```
dandavison, line 33: Does `itemID == ""` correspond to `checkout = true`? As in my other comment, I suggest making checkout a separate operation. (Use `switch` for `actionType`?)

dandavison (commented):
Cool sample! Great idea to have a simple UI as the driver of a sample. Otherwise we're constantly requiring users to make a mental leap from seeing Temporal used from the command line to seeing how it can be used in a real application.

early-return/README.md
```diff
@@ -5,8 +5,15 @@ This sample demonstrates an early-return from a workflow.
...
 a response mid-workflow, while the workflow continues to run to completion.
 
+See [shopping cart](https://github.com/temporalio/samples-go/tree/main/shoppingcart)
+for Update-with-Start being used for lazy initialization.
+
 ### Steps to run this sample:
 1) Run a [Temporal service](https://github.com/temporalio/samples-go/tree/main/#how-to-use).
+
+   NOTE: frontend.enableExecuteMultiOperation=true must be configured for the server
+in order to use Update-with-Start.
```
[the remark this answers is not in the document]
yuandrew: yes! good suggestion

shoppingcart/server/main.go
```diff
@@ -0,0 +1,151 @@
...
+	if err != nil {
+		panic(err)
+	}
+
+	fmt.Println("Starting dummy server...")
+	http.HandleFunc("/", listHandler)
+	http.HandleFunc("/action", actionHandler)
+
+	fmt.Println("Server started on :8080")
+	if err := http.ListenAndServe(":8080", nil); err != nil {
```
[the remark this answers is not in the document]
dandavison: (This isn't blocking, just a suggestion to do at some point)

shoppingcart/README.md
```diff
@@ -0,0 +1,27 @@
...
+    NOTE: frontend.enableExecuteMultiOperation=true must be configured for the server
+in order to use Update-with-Start. 
+
+2) Run the following command to start the worker
+```
+go run shoppingcart/worker/main.go
+```
+3) Run the following command to start the web app
+```
+go run shoppingcart/server/main.go
```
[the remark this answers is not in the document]
yuandrew: agree, copied over from the expense sample, will update that as well

shoppingcart/server/main.go
```diff
@@ -0,0 +1,151 @@
...
+		fmt.Println("Error starting server:", err)
+	}
+}
+
+func listHandler(w http.ResponseWriter, _ *http.Request) {
+	w.Header().Set("Content-Type", "text/html") // Set the content type to HTML
+	_, _ = fmt.Fprint(w, "<h1>DUMMY SHOPPING WEBSITE</h1>"+
```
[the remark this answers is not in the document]
yuandrew: I would agree it doesn't look great, but this UI has minimal styling and I think it's better with a title than not at all. I do think that using the word "Dummy" makes it look quite unprofessional, that should be changed to something like "Sample"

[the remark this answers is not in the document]
dandavison: Yep it was just the word "dummy" that was bothering me! Maybe "Shopping cart"?


It would be nice also to have a tiny bit of styling -- the browser default table styling is ... something else.

shoppingcart/workflow.go
```diff
@@ -0,0 +1,71 @@
...
+		return err
+	}
+
+	err := workflow.Await(ctx, func() bool { return workflow.GetInfo(ctx).GetContinueAsNewSuggested() || checkout })
+	if err != nil {
+		return err
+	}
+	if workflow.GetInfo(ctx).GetContinueAsNewSuggested() {
+		logger.Info("Continuing as new")
+		return workflow.NewContinueAsNewError(ctx, CartWorkflow)
```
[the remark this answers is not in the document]
yuandrew: Good call! I left a `// NOTE:` for cart state across the CAN boundary, I left it unimplemented because I think the main purpose of using CAN here is to show how it could be used. The note gives options on how you could handle cart state, not sure we need to add a specific implementation, especially since the sample itself should never reach that many events.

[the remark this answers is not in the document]
dandavison: Are you sure it's not worth just adding the `cart` parameter to the workflow? It would be more correct and would avoid needing all this text.

shoppingcart/workflow.go
```diff
@@ -0,0 +1,79 @@
...
+			}
+			return nil
+		},
+	}); err != nil {
+		return err
+	}
+
+	signalChan := workflow.GetSignalChannel(ctx, "checkout")
+
+	err := workflow.Await(ctx, func() bool { return workflow.GetInfo(ctx).GetContinueAsNewSuggested() || signalChan.Receive(ctx, nil) })
```
dandavison, line 61: The boolean return value of `signalChan.Receive` doesn't have the right semantics for use in an `Await` -- it can return `false` when a message has been received.

yuandrew: good catch, thanks

shoppingcart/ui/main.go
```diff
@@ -0,0 +1,157 @@
...
+	}
+
+	// Generate the HTML after communicating with the Temporal workflow.
+	// "list" already generates HTML, so skip for that scenario
+	if actionType != "list" {
+		listHandler(w, r)
+	}
+}
+
+func updateWithStartCart(actionType string, id string) shoppingcart.CartState {
```
dandavison, line 108: WDYT about renaming this `id` variable to `itemID` everywhere in this file (incl. in the URL params)? That will help readers not get confused with the other IDs (workflow ID, and update ID; especially since workflow ID is simply called `ID` in the `StartWorkflowOptions` struct)

shoppingcart/workflow.go
```diff
@@ -0,0 +1,86 @@
...
+			return workflow.AllHandlersFinished(ctx)
+		})
+		if err != nil {
+			return err
+		}
+		logger.Info("Continuing as new")
+
+		return workflow.NewContinueAsNewError(ctx, CartWorkflow, cart)
+	}
+
```
dandavison, line 83: Maybe log something like "User has checked out, cart workflow exiting" here?

shoppingcart/ui/main.go
```diff
@@ -0,0 +1,157 @@
...
+func actionHandler(w http.ResponseWriter, r *http.Request) {
+	actionType := r.URL.Query().Get("type")
+	switch actionType {
+	case "checkout":
+		err := workflowClient.SignalWorkflow(context.Background(), sessionId, "", "checkout", nil)
+		if err != nil {
+			log.Fatalln("Error signaling checkout:", err)
+		}
+		sessionId = newSession()
+		log.Println("Items checked out and workflow completed, starting new workflow")
```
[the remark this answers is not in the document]
dandavison: Note that the workflow hasn't necessarily received the signal at this point, so it's misleading to say "workflow completed". It would be most realistic to use an update for checkout, so that the client can wait until the checkout process has completed, but I know that would add complexity to the workflow, so just deleting these two lines would work I think.

dandavison (approved):
Nice, I made a few final comments. I'd argue that https://github.com/temporalio/samples-go/pull/379/files#r1943493462 does need to be addressed before merging, but other than that LGTM!

PR #382, opened 2025-01-23
Waiting activity completion after ExecuteActivityCall

edmondop:
...

### DIFF
--- a/choice-exclusive/workflow.go
+++ b/choice-exclusive/workflow.go
@@ -1,6 +1,7 @@
 package choice
 
 import (
+	"fmt"
 	"time"
 
 	"go.temporal.io/sdk/workflow"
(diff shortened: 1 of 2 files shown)


choice-exclusive/workflow.go
```diff
@@ -33,14 +33,14 @@ func ExclusiveChoiceWorkflow(ctx workflow.Context) error {
 	// choose next activity based on order result
 	switch orderChoice {
 	case OrderChoiceApple:
-		workflow.ExecuteActivity(ctx, orderActivities.OrderApple, orderChoice)
+		workflow.ExecuteActivity(ctx, orderActivities.OrderApple, orderChoice).Get(ctx, nil)
```
cretz, line 36: This is a good change, but we need to handle the errors coming back from these

cretz (approved):
Thanks!

PR #383, opened 2025-01-24
Add retry mechanic to worker-specific-task-queue sample

### LINKED ISSUE #376: [Feature Request] Demonstrate retries in worker-specific-task-queue sample

Users need to see how to retry whole set of things in the sample if, say, worker crashes

yuandrew:
## What was changed

Added an example of a retry mechanic to worker-specific-task-queue

## Why?

Common scenario we want an example for.

## Checklist

1. Closes #376

2. How was this tested:

Tested locally, after running a worker and the starter code. If worker goes down 5 times in a row during an activity and ScheduleToClose window passes, worker prints "Workflow failed after multiple retries" and starter code errors out. 

Added unit tests demonstrating retry mechanic

3. Any docs updates needed?


worker-specific-task-queues/workflow.go
```diff
@@ -11,6 +11,18 @@ import (
 // FileProcessingWorkflow is a workflow that uses Worker-specific Task Queues to run multiple Activities on a consistent
 // host.
 func FileProcessingWorkflow(ctx workflow.Context) (err error) {
+	for attempt := 1; attempt <= 5; attempt++ {
```
cretz, line 14: Don't have to, but may be worth a little comment here on why there is a loop around this whole "process".

worker-specific-task-queues/workflow.go
```diff
@@ -11,6 +11,18 @@ import (
...
+		if err = processFile(ctx); err == nil {
```
cretz, line 15: Arguably one might only retry on schedule to close timeout, but this is probably fine too.

cretz (approved):
Nothing blocking, but would like @Quinn-With-Two-Ns to look

worker-specific-task-queues/README.md
```diff
@@ -28,3 +28,8 @@ Start the Workflow Execution:
 ```bash
 go run worker-specific-task-queues/starter/main.go
 ```
+
+### Things to try
+You can try to intentionally crash Workers while they are doing work to see what happens when work gets "stuck" in a unique queue: currently the Workflow will `scheduleToCloseTimeout` without a Worker, and retry when a Worker comes back online.
+
+After the 5th attempt, it logs `Workflow failed after multiple session retries.` and exits. But you may wish to implement compensatory logic, including notifying you.
```
Quinn-With-Two-Ns, line 35: It doesn't log `Workflow failed after multiple session retries.` it logs `Workflow failed after multiple retries.` no?

worker-specific-task-queues/workflow.go
Quinn-With-Two-Ns, line 14: nit: could use range over int here

worker-specific-task-queues/README.md
```diff
@@ -28,3 +28,8 @@ Start the Workflow Execution:
 ```bash
 go run worker-specific-task-queues/starter/main.go
 ```
+
+### Things to try
```
Quinn-With-Two-Ns, line 32: Do you think we can add a unit test to show the retry working? Not blocking but would be nice to show in a unit test

Quinn-With-Two-Ns (approved):

worker-specific-task-queues/README.md
yuandrew: yep, added!

PR #384, opened 2025-01-27
Add sleep-for-days sample

THardy98:
This is a simple sample demonstrating Temporal's ability to sleep for days and still execute reliably.

##What was changed
added sleep-for-days sample

Any docs updates needed?
I believe so @webchick

### DIFF
--- a/sleep-for-days/sleepfordays_workflow.go
+++ b/sleep-for-days/sleepfordays_workflow.go
@@ -1,4 +1,4 @@
-package helloworld
+package sleepfordays
 
 import (
 	"context"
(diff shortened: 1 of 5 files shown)


sleep-for-days/sleepfordays_workflow.go
```diff
@@ -0,0 +1,37 @@
...
+		StartToCloseTimeout: 10 * time.Second,
+	})
+
+	isComplete := false
+	sigChan := workflow.GetSignalChannel(ctx, CompleteSignal)
+	workflow.Go(ctx, func(ctx workflow.Context) {
+		sigChan.Receive(ctx, &isComplete)
+	})
+
+	for !isComplete {
```
Quinn-With-Two-Ns, line 25: Might be simpler to just check the signal channel every loop, also this sample is a bit different then the python one https://github.com/temporalio/samples-python/commit/2f3f2ac03d75e854e592b476354c0d692eb23f3b. In the python one your signal takes no input, but the Go one does. I think we should match python here.

THardy98: Removed the signal input (and workflow input).

Quinn-With-Two-Ns (approved):

PR #386, opened 2025-01-31
Add missing argument to CAN call

dandavison:
It looks like Go's weak typing was letting us down here and allowing an incorrect call to get merged. Ideally we'd have a test that exercised the CAN.


Quinn-With-Two-Ns:
Ah good catch, yeah we should add a test for CAN.

Quinn-With-Two-Ns (approved):

PR #387, opened 2025-02-02
feat: api key support, hello world example

gauravthadani:
## What was changed

replicated the hello world example, but authenticated with api key

## Why?

Found myself needing a Temporal client with api key and didn't find an example here, so adding 1

## Checklist

1. Closes 

2. How was this tested:

Ran the Workflow against my namespace in Temporal Cloud

3. Any docs updates needed?

Added relevant readme to the sample

### DIFF
--- a/helloworld-apiKey/helloworld.go
+++ b/helloworld-apiKey/helloworld.go
@@ -1,0 +1,92 @@
+package helloworldapiKey
+
+import (
+	"context"
+	"crypto/tls"
+	"flag"
+	"fmt"
+	"os"
+	"time"
+
+	"go.temporal.io/sdk/activity"
+	"go.temporal.io/sdk/client"
+	"go.temporal.io/sdk/workflow"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/metadata"
+)
+
+// Workflow is a Hello World workflow definition.
+func Workflow(ctx workflow.Context, name string) (string, error) {
+	ao := workflow.ActivityOptions{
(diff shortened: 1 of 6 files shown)


Quinn-With-Two-Ns:
Thanks for the sample! Can you please add this to the [main readme](https://github.com/temporalio/samples-go/blob/main/README.md) as well

gauravthadani:
> Thanks for the sample! Can you please add this to the [main readme](https://github.com/temporalio/samples-go/blob/main/README.md) as well

added to README.md , thanks for checking @Quinn-With-Two-Ns

Quinn-With-Two-Ns (approved):

helloworld-apiKey/helloworld.go
```diff
@@ -0,0 +1,92 @@
...
+						metadata.AppendToOutgoingContext(ctx, "temporal-namespace", *namespace),
+						method,
+						req,
+						reply,
+						cc,
+						opts...,
+					)
+				},
+			),
+		},
```
cretz, line 82: Note, this shouldn't be necessary once https://github.com/temporalio/sdk-go/pull/1467 is merged (don't have to change anything now)

Quinn-With-Two-Ns: Not merged, released

gauravthadani: okay, thanks @cretz and @Quinn-With-Two-Ns . merging this for now. 

have a note to revisit once the new version is released 👌

PR #389, opened 2025-02-11
Upgrade Go SDK and update Nexus sample

bergundy:
Also upgrade the API dependency in the grpc-proxy sample

~~**DO NOT MERGE** - this is still pending the Go SDK release.~~


Quinn-With-Two-Ns (approved):

nexus/handler/app.go
```diff
@@ -15,10 +15,11 @@ import (
 )
 
 // NewSyncOperation is a meant for exposing simple RPC handlers.
-var EchoOperation = temporalnexus.NewSyncOperation(service.EchoOperationName, func(ctx context.Context, c client.Client, input service.EchoInput, options nexus.StartOperationOptions) (service.EchoOutput, error) {
+var EchoOperation = nexus.NewSyncOperation(service.EchoOperationName, func(ctx context.Context, input service.EchoInput, options nexus.StartOperationOptions) (service.EchoOutput, error) {
 	// The method is provided with an SDK client that can be used for arbitrary calls such as signaling, querying,
```
Quinn-With-Two-Ns, line 19: This comment no longer makes sense since the method does not provide a client

nexus/README.md
```diff
@@ -24,16 +24,14 @@ This sample shows how to use Temporal for authoring a Nexus service and call it
    site](https://learn.temporal.io/getting_started/go/dev_environment/#set-up-a-local-temporal-service-for-development-with-temporal-cli)
    to install Temporal CLI.
 
-> NOTE: Required version is at least v1.1.0.
```
Quinn-With-Two-Ns, line 27: Do we want to recommend `v1.3.0` (I am assuming that will be the latest release and include all the GA changes)

bergundy: I was debating that but didn't want to be too strict. Rethinking that I think it's probably the best to recommend the latest and greatest.

Quinn-With-Two-Ns: I think if we merge https://github.com/temporalio/sdk-go/pull/1833 we should only be recommending a server release that understands this error shouldn't be retryable

bergundy: Updated the recommendation.

PR #393, opened 2025-02-27
Add nexus-cancelation sample

bergundy:
See also https://github.com/temporalio/samples-java/pull/718

### DIFF
--- a/nexus-cancelation/caller/workflows.go
+++ b/nexus-cancelation/caller/workflows.go
@@ -1,0 +1,75 @@
+package caller
+
+import (
+	"errors"
+
+	"github.com/temporalio/samples-go/nexus/service"
+	"go.temporal.io/sdk/temporal"
+	"go.temporal.io/sdk/workflow"
+)
+
+const (
+	TaskQueue    = "my-caller-workflow-task-queue"
+	endpointName = "my-nexus-endpoint-name"
+)
+
+var languages = []service.Language{
+	service.EN,
+	service.FR,
+	service.DE,
+	service.ES,
+	service.TR,
+}
+
+type result struct {
+	lang   service.Language
+	output service.HelloOutput
+	err    error
+}
+
+func HelloCallerWorkflow(ctx workflow.Context, name string) (string, error) {
+	log := workflow.GetLogger(ctx)
+	c := workflow.NewNexusClient(endpointName, service.HelloServiceName)
+	results := make([]result, len(languages))
+	callCtx, cancel := workflow.WithCancel(ctx)
+	defer cancel()
+
+	// Concurrently execute an operation per language.
+	wg := workflow.NewWaitGroup(ctx)
+	for i, lang := range languages {
+		wg.Add(1)
+		workflow.Go(ctx, func(ctx workflow.Context) {
+			defer wg.Done()
+			// Use the cancelable callCtx for executing the operation.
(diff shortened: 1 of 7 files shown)


nexus-cancelation/caller/workflows.go
```diff
@@ -0,0 +1,75 @@
...
+
+	// Concurrently execute an operation per language.
+	wg := workflow.NewWaitGroup(ctx)
+	for i, lang := range languages {
+		wg.Add(1)
+		workflow.Go(ctx, func(ctx workflow.Context) {
+			defer wg.Done()
+			// Use the cancelable callCtx for executing the operation.
+			fut := c.ExecuteOperation(
+				callCtx,
```
Quinn-With-Two-Ns, line 45:
```suggestion
				ctx,
```

nexus-cancelation/caller/workflows.go
```diff
@@ -0,0 +1,75 @@
...
+				workflow.NexusOperationOptions{})
+			var output service.HelloOutput
+			err := fut.Get(ctx, &output)
+			results[i] = result{
+				lang:   lang,
+				output: output,
+				err:    err,
+			}
+			// The first operation to win the race cancels the others.
+			cancel()
```
Quinn-With-Two-Ns, line 57: This is a bit different then the Java sample, there we waited for everything to start before cancelling

bergundy: That doesn't matter though, since the server will wait for start for us.
You can do the same in Java.

nexus-cancelation/caller/workflows.go
```diff
@@ -0,0 +1,75 @@
...
+	c := workflow.NewNexusClient(endpointName, service.HelloServiceName)
+	results := make([]result, len(languages))
+	callCtx, cancel := workflow.WithCancel(ctx)
+	defer cancel()
+
+	// Concurrently execute an operation per language.
+	wg := workflow.NewWaitGroup(ctx)
+	for i, lang := range languages {
+		wg.Add(1)
+		workflow.Go(ctx, func(ctx workflow.Context) {
```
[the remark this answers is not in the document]
bergundy: Ah good, call, missed this during refactoring, I'm ignoring this sub-context in the call. I wonder if this matters in practice, but agree this is better in a sample.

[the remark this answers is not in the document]
Quinn-With-Two-Ns: Yeah it can, in this case probably not but it is best practice. We always recommend in the Go SDK to use the context of the function you are in.

nexus-cancelation/caller/workflows.go
Quinn-With-Two-Ns: I don't think the sample should be written assuming a generic async operation and not rely on specific behaviour of a workflow run

Quinn-With-Two-Ns (approved):

nexus-cancelation/caller/workflows.go
bergundy: The server doesn't know this is a workflow run operation, just that it's async (as indicated by the token returned from the handler).

bergundy:
I'm going to add a comment to note that the operation must be async in order to be cancelable and merge.

bergundy:
I noticed I haven't added the sample to the README list, I'll add it in the next PR.

PR #394, opened 2025-02-28
Add Nexus context propagation sample

bergundy:
Also added the Nexus Cancelation sample to the README list.
I made some slight adjustments to the `ctxpropgation` since I reused the propagator in this new sample.

### DIFF
--- a/nexus-context-propagation/interceptor.go
+++ b/nexus-context-propagation/interceptor.go
@@ -1,0 +1,114 @@
+package nexuscontextpropagation
+
+import (
+	"context"
+	"encoding/base64"
+	"fmt"
+
+	"github.com/nexus-rpc/sdk-go/nexus"
+	"github.com/temporalio/samples-go/ctxpropagation"
+	"go.temporal.io/api/common/v1"
+	"go.temporal.io/sdk/converter"
+	"go.temporal.io/sdk/interceptor"
+	"go.temporal.io/sdk/workflow"
+	"google.golang.org/protobuf/encoding/protojson"
+)
+
+type WorkerInterceptor struct {
+	interceptor.WorkerInterceptorBase
+	DataConverter converter.DataConverter
+}
+
+func (w *WorkerInterceptor) InterceptWorkflow(ctx workflow.Context, next interceptor.WorkflowInboundInterceptor) interceptor.WorkflowInboundInterceptor {
+	in := &workflowInboundInterceptor{parent: w}
+	in.Next = next
+	return in
+}
+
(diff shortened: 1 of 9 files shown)


nexus-context-propagation/interceptor.go
```diff
@@ -0,0 +1,114 @@
...
+	"github.com/nexus-rpc/sdk-go/nexus"
+	"github.com/temporalio/samples-go/ctxpropagation"
+	"go.temporal.io/api/common/v1"
+	"go.temporal.io/sdk/converter"
+	"go.temporal.io/sdk/interceptor"
+	"go.temporal.io/sdk/workflow"
+	"google.golang.org/protobuf/encoding/protojson"
+)
+
+type WorkerInterceptor struct {
```
Quinn-With-Two-Ns, line 17: I am doing this in Java right now and one thing I debated was just putting all the context propagation in the interceptor. Is is generally what we recommend to customers now

bergundy: Why do we recommend that? I wouldn't expect users to have to know every method that context might be propagated from. I think newer SDKs should implement a context propagation interceptor that could potentially include Nexus if we have a good way to ensure a consistent format (e.g. everything is a payload that can be encrypted).

bergundy: > SDKs could provide a context propagation interceptor helper like we do for tracing

Definitely agree with this. While it's simple to implement one, you still need to know which contexts to propagate to and from and remember to update your interceptors for new APIs that require interception.

nexus-context-propagation/interceptor.go
```diff
@@ -0,0 +1,114 @@
...
+	return n
+}
+
+// ExecuteNexusOperation implements interceptor.WorkflowOutboundInterceptor. It extracts values from workflow context
+// and propagates them via a Nexus header.
+func (out *workflowOutboundInterceptor) ExecuteNexusOperation(
+	ctx workflow.Context,
+	input interceptor.ExecuteNexusOperationInput,
+) workflow.NexusOperationFuture {
+	if values, ok := ctx.Value(ctxpropagation.PropagateKey).(ctxpropagation.Values); ok {
```
Quinn-With-Two-Ns, line 71: In Java I can just copy the headers from the context directly because `MDC` only supports strings, this is fine though for Go

Quinn-With-Two-Ns:
You you could add a sync handler and show the context is propagated there as well

Quinn-With-Two-Ns (approved):

bergundy:
> You you could add a sync handler and show the context is propagated there as well. I understand that your code might implicitly show that, but it might not be obvious to users

I was debating whether to do that or not. Let's merge since this isn't blocking and I think I should have time to do a quick follow up today.

PR #395, opened 2025-02-28
Nexus SyncOperation context propagation

bergundy:


nexus-context-propagation/handler/app.go
```diff
@@ -14,9 +14,26 @@ import (
 	"github.com/temporalio/samples-go/nexus/service"
 )
 
+// NewSyncOperation is a meant for exposing simple RPC handlers.
+var EchoOperation = nexus.NewSyncOperation(service.EchoOperationName, func(ctx context.Context, input service.EchoInput, options nexus.StartOperationOptions) (service.EchoOutput, error) {
+	// Values may be extracted from the context in the Operation handler body.
+	values, ok := ctx.Value(ctxpropagation.PropagateKey).(ctxpropagation.Values)
+	if ok {
+		input.Message += ", " + values.Key + ": " + values.Value
```
Quinn-With-Two-Ns, line 22: We could just log the context value, we generally using headers to pass input

bergundy: I don't have a strong opinion, do you think it's worth blocking the PR for this?

Quinn-With-Two-Ns: Are samples should reflect what we recommend users do in production. Maxim has called out our samples that do things we wouldn't recommend users do in production so yes I think we should change.

PR #396, opened 2025-03-03
Nexus minor README update

yuandrew:
## What was changed

Removed Public preview line now that nexus is GA.

Fixed some typos, added a feedback point on the nexus sample

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/nexus/README.md
+++ b/nexus/README.md
@@ -60,7 +60,7 @@ temporal operator nexus endpoint create \
 ## Getting started with a self-hosted service or Temporal Cloud
 
 Nexus is currently available as
-[Public Preview](https://docs.temporal.io/evaluate/development-production-features/release-stages).
+[General Availability](https://docs.temporal.io/evaluate/development-production-features/release-stages).
 
 Self hosted users can [try Nexus
 out](https://github.com/temporalio/temporal/blob/main/docs/architecture/nexus.md#trying-nexus-out) in single cluster
--- a/nexus/README.md
+++ b/nexus/README.md
@@ -60,7 +60,7 @@ temporal operator nexus endpoint create \
 ## Getting started with a self-hosted service or Temporal Cloud
 
 Nexus is currently available as
-[General Availability](https://docs.temporal.io/evaluate/development-production-features/release-stages).
+[General Availability](https://docs.temporal.io/evaluate/development-production-features/release-stages#general-availability).
 
 Self hosted users can [try Nexus
 out](https://github.com/temporalio/temporal/blob/main/docs/architecture/nexus.md#trying-nexus-out) in single cluster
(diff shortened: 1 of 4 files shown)


nexus/caller/workflows.go
```diff
@@ -32,6 +32,7 @@ func HelloCallerWorkflow(ctx workflow.Context, name string, language service.Lan
 
 	// Optionally wait for the operation to be started. NexusOperationExecution will contain the operation token in
 	// case this operation is asynchronous.
+	// TODO: Is it worth explaining what you can do with the nexus operation token?
```
yuandrew, line 35: @prasek small feedback on this sample, it's mentioned that there's an operation token if the operation is async, but as someone learning about nexus, I don't know why that's important or what its used for. I think if we mention the operation token here, we should give a little more information on its use

nexus/README.md
```diff
@@ -60,7 +60,7 @@ temporal operator nexus endpoint create \
 ## Getting started with a self-hosted service or Temporal Cloud
 
 Nexus is currently available as
-[Public Preview](https://docs.temporal.io/evaluate/development-production-features/release-stages).
+[General Availability](https://docs.temporal.io/evaluate/development-production-features/release-stages#general-availability).
```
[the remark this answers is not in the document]
yuandrew: makes sense

cretz (approved):

nexus/caller/workflows.go
Quinn-With-Two-Ns: Just to note, please remove the TODO before merging

Quinn-With-Two-Ns: I do agree we should explain or link to an explanation of the token

prasek: @yuandrew thanks agree we should include more here from the content in https://docs.temporal.io/nexus/operations of note

> Unlike a traditional RPC, an asynchronous Nexus Operation has an operation token that can be used to re-attach to a long-lived Nexus Operation, for example, one backed by a Temporal Workflow. Nexus Operations support a uniform interface to get the status of an operation or its result, receive a completion callback, or cancel the operation – all of which are fully integrated into the Temporal Platform.

> [Asynchronous Nexus Operations](https://docs.temporal.io/nexus/operations#asynchronous-operation-lifecycle), created with the New-Workflow-Run-Operation SDK helper, will return a Nexus Operation token, that can be used to perform additional actions like canceling an Operation.

nexus/caller/workflows.go
```diff
@@ -32,6 +32,7 @@ func HelloCallerWorkflow(ctx workflow.Context, name string, language service.Lan
 
 	// Optionally wait for the operation to be started. NexusOperationExecution will contain the operation token in
 	// case this operation is asynchronous.
```
prasek, line 34:
```suggestion
	// case this operation is asynchronous, which is a handle that can be used to perform additional actions like cancelling an operation.
```

PR #399, opened 2025-04-04
Update stale doc links in README.md

lilydoar:
## What was changed
Fix stale link to signal documentation

## Why?
One of the doc links pointed to a page which no longer exists


cretz (approved):
Thanks!

PR #401, opened 2025-04-19
Add nde example

Ozoniuss:
This PR is not ready -- I'm looking to gather some feedback on whether you like the format of this example. I'm going to extend this to cover multiple cases (e.g. branching logic, changing an Activity's argument type etc.) if you like the format.

## What was changed
Adds examples for how to obtain NDEs with Temporal. I will update this section further once the examples are complete.

## Why?
There is no explicit example on which actions can run into NDEs. Having those examples would be greatly beneficial for developers wondering if their change leads to an NDE. For example, I wasn't able to find in the docs if changing an Activity name (e.g. as part of a refactor) will lead to an NDE. The docs state

```
If a corresponding Event already exists within the Event History that maps to the generation of that Command in the same sequence, and some specific metadata of that Command matches with some specific metadata of the Event, then the Function Execution progresses.
[...]
If a generated Command doesn't match what it needs to in the existing Event History, then the Workflow Execution returns a non-deterministic error.
```
but it's not obvious what "needs to match". Obviously compiling an exhaustive list of all scenarios and keeping it up to date is probably not desirable, but I wanted to have a list of samples that cover common development use cases.
 
## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/nondeterminism/README.md
+++ b/nondeterminism/README.md
@@ -1,0 +1,17 @@
+# Non-deterministic error examples
+
+These samples showcase various cases in which you can run into non-deterministic errors. The steps explaining how to run the code such that such error is thrown are explained in each workflow sample.
+
+See [the deterministic constraints](https://docs.temporal.io/workflow-definition#deterministic-constraints) as a docs reference.
+
+### Steps to run this sample:
+1) Run a [Temporal service](https://github.com/temporalio/samples-go/tree/main/#how-to-use).
+2) Run the following command to start the worker
+```
+go run nondeterminism/worker/main.go
+```
+3) Run the following command to start the example
+```
+go run nondeterminism/starter/main.go
+```
+4) Check the instruction for each individual workflow to trigger the NDE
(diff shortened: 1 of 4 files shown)


Ozoniuss:
Okay, thanks for letting me know. For me, I find the sample to be useful because
- it makes it easy to understand exactly which changes I shouldn't make;
- I like having the code to be able to test it and explore the nde behaviour;
- I pretty much use this repo all the time when working with new concepts/ revisiting concepts I already learned so it was simpler to have it there.

However, if you prefer changes to the docs, I'm happy to put these separately in one of my personal gists/repos and look into potentially improving the docs.

cretz:
Yeah I believe this fits docs/education better than these samples (but the docs/education content can of course have code demonstrating). Feel free to open an issue on https://github.com/temporalio/documentation.

cretz:
(closing in favor of documentation/education route)

PR #404, opened 2025-05-06
Add dynamic workflow sample

yuandrew:
## What was changed

Added a sample for using dynamic workflows and activities

## Why?

new Go feature

## Checklist

1. Closes 

2. How was this tested:

Ran sample with https://github.com/temporalio/sdk-go/pull/1946 locally

3. Any docs updates needed?


Quinn-With-Two-Ns:
nit: we should also cover how to test a dynamic workflow and activity with the test suite

Quinn-With-Two-Ns:
@yuandrew can we merge this now?

Quinn-With-Two-Ns (approved):

PR #412, opened 2025-06-19
Update polling sample to use benign exceptions

Quinn-With-Two-Ns:
Update polling sample to use benign exceptions


Quinn-With-Two-Ns:
Note: This can't be merged until the Go SDK is released

THardy98 (approved):

PR #416, opened 2025-07-07
Bump SDK version to 1.35.0

yuandrew:
## What was changed

Bump SDK version to 1.35.0, then `go mod tidy`

## Why?

new SDK release

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


go.mod
```diff
@@ -86,12 +88,12 @@ require (
...
 	go4.org/intern v0.0.0-20230525184215-6c62f75575cb // indirect
 	go4.org/unsafe/assume-no-moving-gc v0.0.0-20231121144256-b99613f794b6 // indirect
-	golang.org/x/crypto v0.27.0 // indirect
+	golang.org/x/crypto v0.37.0 // indirect
 	golang.org/x/mod v0.21.0 // indirect
-	golang.org/x/net v0.29.0 // indirect
-	golang.org/x/sync v0.8.0 // indirect
-	golang.org/x/sys v0.25.0 // indirect
-	golang.org/x/text v0.18.0 // indirect
+	golang.org/x/net v0.39.0 // indirect
```
Quinn-With-Two-Ns, line 93: This version may be to high if you are trying to build with Go version `1.23`

yuandrew: How come that would be too high? The `go.mod` for that version shows 1.23.0 https://cs.opensource.google/go/x/net/+/refs/tags/v0.39.0:go.mod

Quinn-With-Two-Ns (approved):

PR #418, opened 2025-07-28
Fix .devcontainer

derwasp:
## Motivation
The existing .devcontainer config contains a missmatched version of go as well as it's using an outdated temporal toolset.

## Proposed changes
* Remove docker-compose
* Remove custom dockerfile
* Use the official devcontainers: https://github.com/devcontainers/images/tree/main/src/go
* Add temporal CLI installation


derwasp:
> Is the current setup broken?

Yep, when I tried running it, the samples didn't compile.
I can't just update the image version, since the one that was referenced in the Dockerfile is deprecated. The whole [repo was archived](https://github.com/microsoft/vscode-dev-containers/blob/v0.231.6/containers/go/.devcontainer/base.Dockerfile) in 2023.

My understanding is that you used the "all of the custom docker setup, i.e. the postgressql db." to run the non-cli version of Temporal? Is this not correct? If yes, the new setup just installs the temporal cli in the `postCreateCommand`, hence I assume you don't quite need all the other components.

PR #420, opened 2025-07-31
External environment configuration sample

### LINKED ISSUE #390: [Feature Request] External client configuration

Update all samples to leverage external client configuration once https://github.com/temporalio/sdk-go/issues/1815 is complete.

yuandrew:
## What was changed

Added example using external environment configuration

## Why?

new feature!

## Checklist

1. Closes #390 

2. How was this tested:
Ran locally, some methods don't "work" out of the box, but demonstrate how users would use this new feature

3. Any docs updates needed?

### DIFF
--- a/external-env-conf/README.md
+++ b/external-env-conf/README.md
@@ -1,0 +1,29 @@
+# Temporal External Client Configuration Samples
+
+This directory contains samples that demonstrate how to use the external client configuration feature.
+This feature allows you to configure a client using a TOML file and/or environment variables,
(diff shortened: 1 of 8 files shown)


README.md
cretz, line 1: This does not close #390. The issue there is to update _every_ sample to using external environment configuration instead of assuming localhost default so _every_ sample can use environment configuration. I'm not even sure we really need an environment configuration sample, but I have no problem with it.

PR #424, opened 2025-09-05
Search attribute sample facelift

bergundy:
## What was changed

- Update the search attributes sample to use typed search attributes.
- Showed how to start the dev server with search attributes.
- Used `activity.GetClient` instead of the old method of passing client via context.

## Why?

I wanted to see a sample for typed search attributes and couldn't find one, then realized it needed a facelift.


Quinn-With-Two-Ns:
There is already a sample for [typed search attributes](https://github.com/temporalio/samples-go/tree/main/typed-searchattributes) if you want to make any improvements please make them there. 

This sample should be deleted.

Quinn-With-Two-Ns (changes requested):
> I wanted to see a sample for typed search attributes and couldn't find one, then realized it needed a facelift.

https://github.com/temporalio/samples-go/tree/main/typed-searchattributes

Feel free to delete this one and rename the other one. Originally we left it here since users using old SDKs would still get value, but it has probably been long enough now.

bergundy:
@Quinn-With-Two-Ns should we just replace the previous `typed-searchattributes` sample with this one?

Quinn-With-Two-Ns:
Yeah that works for me, as I said [above](https://github.com/temporalio/samples-go/pull/424#pullrequestreview-3190502017) we had some reason for leaving this sample, but I think typed search attributes have been around long enough

Quinn-With-Two-Ns:
Thanks! we just need to update the top level `README` as well to remove the reference to the now deleted sample

bergundy:
Done.

PR #442, opened 2026-01-08
Payload converter based on payload size

Ozoniuss:
This example shows how to implement a payload converter that detects when a payload is larger than a certain threshold, in which case it writes it to a storage system of choice instead of the temporal server. In this particular example payloads are written to a local file but it can be adapted to work with any storage. Such a converter is useful when a workflow/activity regularly has payloads that risk exceeding Temporal's hard limit of 2MB.

## What was changed
Added a new sample.

## Why?
Two places where I worked required building something similar, so I thought it would be a good idea to write a sample for it. I haven't seen a similar sample available.

## Checklist

1. How was this tested:
Locally with Temporal cluster as well as unit tests available.

2. Any docs updates needed?
No

### DIFF
--- a/data-converter-large-payloads/README.md
+++ b/data-converter-large-payloads/README.md
@@ -1,0 +1,27 @@
+This sample shows an implementation of a payload converter that automatically detects workflow and activity payloads larger than a certain threshold and writes them to a file if that's the case.
+
+### Steps to run this sample:
+1) Run a [Temporal service](https://github.com/temporalio/samples-go/tree/main/#how-to-use).
+2) Run the following command to start the worker
+```
+go run data-converter-large-payloads/worker/main.go
+```
+3) Run the following command to start the example
+```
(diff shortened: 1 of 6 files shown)


data-converter-large-payloads/README.md
cretz: If you are going to do implicit large payload offloading, we recommend using a codec instead of a converter, as https://github.com/DataDog/temporal-large-payload-codec has.

However, with that third party example, we are not necessarily looking for a blessed example of this at this time for two reasons. First, an ideal sample since Temporal is distributed would not write to disk. Second, this isn't often a preferred pattern as arbitrarily offloading large data to/from external stores can hide from the workflow authors that they should consider being more explicit about offloading large data instead of constantly doing this on replay and such. Finally, we are actively working on improving this situation soon doing basically this exact thing but providing more explicit external payload storage interfaces and warnings.

PR #456, opened 2026-04-13
Add sample for stand alone nexus operations

Quinn-With-Two-Ns:
Sample for stand alone nexus operations


nexus-standalone-operations/handler_test.go
```diff
@@ -0,0 +1,183 @@
...
+// operations on the dev server. These mirror the flags used in the SDK's own
+// integration tests (sdk-go/internal/cmd/build/main.go:152-162).
+func devServerExtraArgs() []string {
+	return []string{
+		"--http-port", "7243",
+		// TODO Use correct DC values once the dev server build supporting standalone Nexus is released.
+		"--dynamic-config-value", `component.callbacks.allowedAddresses=[{"Pattern":"*","AllowInsecure":true}]`,
+		"--dynamic-config-value", `system.refreshNexusEndpointsMinWait="0s"`,
+		"--dynamic-config-value", "component.nexusoperations.recordCancelRequestCompletionEvents=true",
+		"--dynamic-config-value", "history.enableRequestIdRefLinks=true",
```
bergundy, line 40: We should make this true by default if it isn't already.

nexus-standalone-operations/handler_test.go
```diff
@@ -0,0 +1,183 @@
...
+		"--dynamic-config-value", "history.enableChasm=true",
+		"--dynamic-config-value", "component.nexusoperations.useSystemCallbackURL=false",
```
bergundy, line 42: This will be true by default in 1.31 (first release to support SANO). For now set it to true please.

nexus-standalone-operations/handler_test.go
```diff
@@ -0,0 +1,183 @@
...
+		"--dynamic-config-value", `component.nexusoperations.callback.endpoint.template="http://localhost:7243/namespaces/{{.NamespaceName}}/nexus/callback"`,
```
bergundy, line 43: Not required anymore.

nexus-standalone-operations/handler_test.go
```diff
@@ -0,0 +1,183 @@
...
+// devServerExtraArgs returns the CLI flags needed to enable standalone Nexus
+// operations on the dev server. These mirror the flags used in the SDK's own
+// integration tests (sdk-go/internal/cmd/build/main.go:152-162).
+func devServerExtraArgs() []string {
+	return []string{
+		"--http-port", "7243",
+		// TODO Use correct DC values once the dev server build supporting standalone Nexus is released.
+		"--dynamic-config-value", `component.callbacks.allowedAddresses=[{"Pattern":"*","AllowInsecure":true}]`,
+		"--dynamic-config-value", `system.refreshNexusEndpointsMinWait="0s"`,
+		"--dynamic-config-value", "component.nexusoperations.recordCancelRequestCompletionEvents=true",
```
bergundy, line 39: Should default this to true if it isn't already.

nexus-standalone-operations/handler_test.go
```diff
@@ -0,0 +1,183 @@
...
+	taskQueue    = "nexus-standalone-operations-test"
+	endpointName = "nexus-standalone-operations-test-endpoint"
+)
+
+// devServerExtraArgs returns the CLI flags needed to enable standalone Nexus
+// operations on the dev server. These mirror the flags used in the SDK's own
+// integration tests (sdk-go/internal/cmd/build/main.go:152-162).
+func devServerExtraArgs() []string {
+	return []string{
+		"--http-port", "7243",
```
bergundy, line 35: Shouldn't be required.

nexus-standalone-operations/handler_test.go
```diff
@@ -0,0 +1,183 @@
...
+		opName string,
+		input any,
+		options client.StartNexusOperationOptions,
+	) client.NexusOperationHandle {
+		t.Helper()
+		var handle client.NexusOperationHandle
+		require.Eventually(t, func() bool {
+			var execErr error
+			handle, execErr = nexusClient.ExecuteOperation(ctx, opName, input, options)
+			return execErr == nil
```
bergundy, line 116: I think we should fix the need to retry for tests before we launch SANO. It's going to be annoying to use.
You should also only filter out not found errors here.

go.mod
```diff
@@ -4,6 +4,10 @@ go 1.24.0
 
 replace github.com/cactus/go-statsd-client => github.com/cactus/go-statsd-client/v5 v5.0.0
 
+replace go.temporal.io/sdk => /Users/quinnklassen/Documents/Code/sdk-go
```
Evanthx, line 7: I know you said this wasn't ready for checkin, but figured it was worth pointing out these two replace commands!

nexus-standalone-operations/README.md
```diff
@@ -0,0 +1,66 @@
+This sample demonstrates how to use Standalone Nexus Operations (executing Nexus operations directly from client code without wrapping them in a Workflow).
+It shows both sync and async (workflow-backed) operations, and how to use the `ListNexusOperations` and `CountNexusOperations` APIs.
+
+## Note: Standalone Nexus operations require a server version that supports this feature.
```
Evanthx, line 4: Is it worth putting something about how to tell if the server supports the feature? I certainly don't know, so if I run this and it doesn't work, I'll have no idea if the issue is that simple or not!

nexus-standalone-operations/README.md
```diff
@@ -0,0 +1,66 @@
...
+
+2) Create a Nexus endpoint that routes to the worker's task queue. In a second terminal, run:
+```bash
+temporal operator nexus endpoint create \
+  --name nexus-standalone-operations-endpoint \
+  --target-namespace default \
+  --target-task-queue nexus-standalone-operations
+```
+
+3) Open a third terminal, and run the following command to start the worker. The worker is a blocking process that runs until it receives a SIGINT (Ctrl + C) command.
```
Evanthx, line 28: Nitpick, but the second terminal command just runs and is done, so this can be in the same terminal instead of in a third terminal.

nexus-standalone-operations/README.md
```diff
@@ -0,0 +1,66 @@
...
+```bash
+go run nexus-standalone-operations/worker/main.go
+```
+
+You should see two console log lines:
+1. Creating the logger
```
Evanthx, line 34: I did not see this line

nexus-standalone-operations/README.md
```diff
@@ -0,0 +1,66 @@
...
+
+4) In a fourth terminal, run the following command to start the example:
+```bash
+go run nexus-standalone-operations/starter/main.go
+```
+
+You should see something similar to the following output:
+
+```bash
+2026/04/11 14:12:00 INFO  No logger configured for temporal client. Created default one.
```
Evanthx, line 54: I get errors here, but am guessing I don't have something yet? I know this isn't release code yet so is probably related to that.

nexus-standalone-operations/starter/main.go:31:24: c.NewNexusClient undefined (type client.Client has no field or method NewNexusClient)
nexus-standalone-operations/starter/main.go:31:46: undefined: client.NexusClientOptions
nexus-standalone-operations/starter/main.go:40:143: undefined: client.StartNexusOperationOptions
nexus-standalone-operations/starter/main.go:57:168: undefined: client.StartNexusOperationOptions
nexus-standalone-operations/starter/main.go:74:21: c.ListNexusOperations undefined (type client.Client has no field or method ListNexusOperations)
nexus-standalone-operations/starter/main.go:74:70: undefined: client.ListNexusOperationsOptions
nexus-standalone-operations/starter/main.go:91:22: c.CountNexusOperations undefined (type client.Client has no field or method CountNexusOperations)
nexus-standalone-operations/starter/main.go:91:72: undefined: client.CountNexusOperationsOptions

nexus-standalone-operations/starter/main.go
```diff
@@ -0,0 +1,100 @@
...
+)
+
+// This sample demonstrates standalone Nexus operations — executing Nexus operations
+// directly from client code without wrapping them in a workflow.
+
+const endpointName = "nexus-standalone-operations-endpoint"
+
+func main() {
+	// The client is a heavyweight object that should be created once per process.
+	c, err := client.Dial(envconfig.MustLoadDefaultClientOptions())
```
Evanthx, line 23: The other Nexus samples are using "options.ParseClientOptionFlags(os.Args[1:])" before client.Dial, should this as well for consistent behavior?

nexus-standalone-operations/worker/main.go
```diff
@@ -0,0 +1,43 @@
...
+
+	"github.com/temporalio/samples-go/nexus/handler"
+	"github.com/temporalio/samples-go/nexus/service"
+)
+
+const taskQueue = "nexus-standalone-operations"
+
+func main() {
+	// The client and worker are heavyweight objects that should be created once per process.
+	c, err := client.Dial(envconfig.MustLoadDefaultClientOptions())
```
Evanthx, line 21: As above - the other Nexus samples are using "options.ParseClientOptionFlags(os.Args[1:])" before client.Dial, should this as well for consistent behavior?

nexus-standalone-operations/starter/main.go
```diff
@@ -0,0 +1,100 @@
...
+	if err != nil {
+		log.Fatalln("Unable to create client", err)
+	}
+	defer c.Close()
+
+	// Create a NexusClient bound to the endpoint and service.
+	// The endpoint must be pre-created on the server (see README).
+	nexusClient, err := c.NewNexusClient(client.NexusClientOptions{
```
Evanthx, line 31: This is an odd one when looking at other Nexus sample code in here. I went to the equivalent spot and there is no NewNexusClient, so I thought maybe that's new? But then I found it in workflows.go at the top level, which isn't in this sample.

Short version - I am reading the code to see what is different, and so get thrown off - and mostly I was reading this to learn how to use Nexus for standalone operations, and I can't tell what makes this standalone as opposed to self hosted!

nexus-standalone-operations/starter/main.go
```diff
@@ -0,0 +1,100 @@
...
+	log.Println("Started Hello operation", "OperationID", helloHandle.GetID())
+
+	var helloResult service.HelloOutput
+	err = helloHandle.Get(context.Background(), &helloResult)
+	if err != nil {
+		log.Fatalln("Unable to get Hello operation result", err)
+	}
+	log.Println("Hello result:", helloResult.Message)
+
+	// List Nexus operations using the base client (not NexusClient).
```
Evanthx, line 73: I am pretty sure this is just me being new to Nexus still but -- does the reader know what you are talking about here? Why are there two clients? Why would I use the base client? Base implies that the NexusClient has it as a base class, but clearly that's not the case!

Evanthx (commented):
Asking some questions, not sure how many are just me being new still!

nexus-standalone-operations/README.md
Quinn-With-Two-Ns: The samples would fail with a clear `Unsupported error`, but we could certainly document in the readme what to look for

nexus-standalone-operations/starter/main.go
Quinn-With-Two-Ns: Other Nexus samples starter code is starting a workflow not a Nexus operation, the comparison isn't between standalone and self hosted, the comparison is between inside a workflow and outside a workflow

nexus-standalone-operations/starter/main.go
Quinn-With-Two-Ns: >Base implies that the NexusClient has it as a base class

Go has no concept of classes or base class so I don't think that is a conclusion a Go developer would reach, that being said I welcome any suggestions to make the wording hear clearer.

nexus-standalone-operations/handler_test.go
Quinn-With-Two-Ns: Yeah please note the TODO these are just copy pasted from SDK CI not the min. list of flags we will need by the time we release the server

nexus-standalone-operations/README.md
Quinn-With-Two-Ns: When we do know the server version that will support Standalone Nexus operations we will also put that here of course

nexus-standalone-operations/README.md
Quinn-With-Two-Ns: Yeah this can't be run without a bunch of manual work until we get everything merged and released.

nexus-standalone-operations/starter/main.go
Quinn-With-Two-Ns: Yeah bit on the fence about this since the `ParseClientOptionFlags(os.Args[1:])` was only added because we didn't have `envconfig` now that we do we should probably just lean on the since very other sample uses it. For consistency we should probably remove `ParseClientOptionFlags` from other samples and just lean on `envconfig`

nexus-standalone-operations/starter/main.go
Quinn-With-Two-Ns: I decided to use actual types instead of just "base"

nexus-standalone-operations/handler_test.go
```diff
@@ -0,0 +1,161 @@
...
+
+	// Create a standalone NexusClient.
+	nexusClient, err := c.NewNexusClient(client.NexusClientOptions{
+		Endpoint: endpointName,
+		Service:  service.HelloServiceName,
+	})
+	require.NoError(t, err)
+
+	// executeWithRetry retries ExecuteOperation until the endpoint has propagated.
+	// The endpoint registry is eventually consistent.
```
VegetarianOrc: Would be good to have a link to an issue or some other way to track when we can remove this once the server side changes are in.

VegetarianOrc (approved):
LGTM!

jsundai:
@Quinn-With-Two-Ns Is this sample ready to be merged?

tconley1428 (approved):

PR #463, opened 2026-05-01
Upgrade Go SDK to v1.43.0

yuandrew:
## What was changed
Upgrade Go SDK version to v1.43.0

## Why?
New version


yuandrew:
Looks like this was already bumped, closing

PR #474, opened 2026-05-20
External storage and codec server sample

jmaeagle99:
## What was changed

Add sample for external storage with codec server that supports multiple namespaces and downloads.

## Why?

Provide a baseline reference for how to use external storage with the S3 driver and an updated codec server.

## Checklist

1. How was this tested: Locally ran; investigating possibility of adding tests
1. Any docs updates needed? No


external-storage/codec-server/main.go
```diff
@@ -0,0 +1,141 @@
...
+	errCh := make(chan error, 1)
+	go func() { errCh <- srv.ListenAndServe() }()
+
+	fmt.Printf("Codec server running at http://%s, ctrl+c to exit\n", srv.Addr)
+
+	sigCh := make(chan os.Signal, 1)
+	signal.Notify(sigCh, os.Interrupt)
+	select {
+	case <-sigCh:
+		_ = srv.Close()
```
cconstable, line 135: In a production codec server, we would want to drain current requests/tasks before shutting down right? I don't know if we offer guidance on this or it's just a "yea you should know that" kind of thing. I don't have strong feelings about mentioning that here or not because it might be too in the weeds.

external-storage/README.md
```diff
@@ -0,0 +1,133 @@
...
+
+A mock S3 service ([s3-mock](./s3-mock)) is included so you can run the sample
+locally without an AWS account or Docker. A codec server
+([codec-server](./codec-server)) is included to retrieve and decompress payloads
+on demand for the Temporal Web UI.
+
+## Steps to run this sample
+
+1. Run a [Temporal service](https://github.com/temporalio/samples-go/tree/main/#how-to-use).
+2. In a separte terminal, run the mock S3 server. It listens on `:5000` and
```
cconstable: small typo

cconstable (approved):
👍

external-storage/README.md
```diff
@@ -0,0 +1,133 @@
...
+(`BatchSummary`) compress to a few hundred bytes and remain inline.
+
+A mock S3 service ([s3-mock](./s3-mock)) is included so you can run the sample
+locally without an AWS account or Docker. A codec server
+([codec-server](./codec-server)) is included to retrieve and decompress payloads
+on demand for the Temporal Web UI.
+
+## Steps to run this sample
+
+1. Run a [Temporal service](https://github.com/temporalio/samples-go/tree/main/#how-to-use).
```
yuandrew, line 28: nit: It's more verbose, but when I wrote standalone activities sample, I mirrored `helloworld` sample's philosophy of also giving the code and the expected output, like 


> If you used the above instructions to start the server, you should see a line about the CLI, Server and UI versions, and one line each for the Server URL, UI URL and Metrics endpoint. It should look something like this:

```bash
> temporal server start-dev
Temporal CLI 1.7.0 (Server 1.31.0, UI 2.49.1)

Server:  localhost:7233
UI:      http://localhost:8233
Metrics: http://localhost:57058/metrics
```

yuandrew: I don't feel strongly about this, happy to hear pushback. It technically saves users time, but also clicking on 1 extra link is NBD, they can just open it in a new tab and then close it out when they're done. But i guess it's also NBD for us to include the code to run to start the server too 🤷

external-storage/README.md
```diff
@@ -0,0 +1,133 @@
...
+   so multi-namespace setups can dispatch by reading that header.
+
+   | Endpoint | Behavior |
+   | --- | --- |
+   | `POST /encode` | Compress the payload, then offload to S3 if it exceeds the threshold. |
+   | `POST /decode` | Retrieve any external storage references from S3, then decompress. Pass `?preserveStorageRefs=true` to leave references as-is. |
+   | `POST /download` | All inputs must be storage references. Retrieves them from S3 and decompresses. |
+6. Run `temporal workflow show` to see how payloads are stored:
+    ```
+    temporal workflow show --workflow-id external-storage-<timestamp>
```
yuandrew, line 71: Probably worth mentioning explicitly where users get the workflow-id from the output above, (i.e. from step 4's output)

external-storage/README.md
```diff
@@ -0,0 +1,133 @@
...
+        converter.GetDefaultDataConverter(),
+        converter.NewZlibCodec(converter.ZlibCodecOptions{AlwaysEncode: true}),
+    ),
+    ExternalStorage: converter.ExternalStorage{
+        Drivers: []converter.StorageDriver{driver},
+    },
+})
+```
+
+On the encode path the SDK:
```
yuandrew, line 102: nit: I think adding a comment here could make it a little less awkward to read

```suggestion
On the encode path, the SDK:
```

alternatively, a more prose language could be 

```suggestion
When encoding a payload, the SDK:
```

yuandrew: similar comment for decode below

external-storage/README.md
```diff
@@ -0,0 +1,133 @@
...
+4. If still above the threshold, stores the compressed bytes in S3 via
+   the SDK's `s3driver` and replaces the inline payload with a claim-check
+   reference.
+
+On the decode path the SDK reverses these steps, transparently retrieving from
+S3 and decompressing as needed.
+
+The worker, the starter, and the codec server must use the **same** codec and
+external storage configuration so each side can read what the other wrote. In
+this sample the shared wiring lives in
```
yuandrew, line 116:
```suggestion
this sample, the shared wiring lives in
```

external-storage/codec-server/main.go
jmaeagle99: These samples tend to be about how to use features, not necessarily about production best practices. However, in the case of codec servers, it might be warranted since these are our main demonstrations of these types of servers. I can do a separate follow up for this.

external-storage/workflows.go
```diff
@@ -0,0 +1,143 @@
+package externalstorage
```
yuandrew, line 1: this file is usually called `workflow.go`, looks like there's one or 2 other samples that uses `workflows.go`. Should fix these all to be consistent.

external-storage/README.md
```diff
@@ -0,0 +1,133 @@
...
+
+**Scenario:** A fulfillment center processes batches of shipping orders. The
+workflow receives a small request (a batch ID and order count), then internally
+calls a `FetchOrders` activity that returns the full list of orders with
+customer records, line items, and handling notes. That list — several hundred
+kilobytes even after compression — is passed to a second `ProcessOrders`
+activity. Finally the workflow returns a small `BatchSummary` with totals.
+
+Each payload is first compressed by the SDK's `NewZlibCodec`. The SDK then
+checks the compressed size against the default 256 KiB threshold; payloads
```
yuandrew, line 16: It might be helpful to describe explicitly which/how many payloads in the sample will be offloaded to external storage, and how many will go down the regular path. To highlight clarity on thie key feature for this sample

external-storage/README.md
```diff
@@ -0,0 +1,133 @@
...
+   Example output:
+    ```
+    Starting workflow external-storage-20260515-120000 (batch_id=BATCH-20260515-120000, order_count=200)
+
+    Batch BATCH-20260515-120000: 200 orders processed
+      Total shipping cost: $28512.40
+      Total weight:        19684.2 kg
+      Avg delivery:        4.4 days
+    ```
+5. (Optional) Run the codec server in a fourth terminal:
```
yuandrew, line 51: Maybe a silly question, but just want to make sure we don't need to run the codec before running the starter code itself

yuandrew (commented):
It would be good to have a test, most of our samples have a test to go with it, I guess for when we bump the SDK version, we have another sanity check?

external-storage/README.md
jmaeagle99: Correct, not necessary for the starter or worker. Only necessary for the Web UI in this sample. Can add comments to that effect.

external-storage/README.md
jmaeagle99: This new sample strongly resemble the codec-server and encryption samples, which also use the link to how-to-use. I will leave the link in but call out that temporal CLI is easiest for local development and show the expected output.

jmaeagle99: Opened readme update PR at https://github.com/temporalio/samples-go/pull/485

external-storage/workflows.go
jmaeagle99: Open code and test update PR at https://github.com/temporalio/samples-go/pull/484

PR #498, opened 2026-06-30
NEXUS-474: Add sample for Workflow Update Nexus op

mavemuri:
## What was changed

Added sample for Workflow Update Nexus op
## Why?

Snippet to show usage 
## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

- [ ]


go.mod
```diff
@@ -1,9 +1,11 @@
 module github.com/temporalio/samples-go
 
-go 1.25.0
+go 1.25.4
 
 replace github.com/cactus/go-statsd-client => github.com/cactus/go-statsd-client/v5 v5.0.0
 
+replace go.temporal.io/sdk => ../sdk-go
```
mavemuri: uprev after https://github.com/temporalio/sdk-go/pull/2417

nexus-operations/update-workflow/caller/workflow.go
```diff
@@ -0,0 +1,29 @@
...
+	"go.temporal.io/sdk/workflow"
+)
+
+const (
+	TaskQueue    = "remote-update-caller-tq"
+	endpointName = "counter-update-endpoint"
+)
+
+// Executes the UpdateWorkflow on the handler namespace's workflow via Nexus Operation
+func UpdateRemoteCounterWorkflow(ctx workflow.Context, input api.Input) (api.Output, error) {
```
Quinn-With-Two-Ns, line 13: It is debatable, but I would lean towards the caller using a Standalone Nexus operation because it make the sample way simpler where you just need the start and the handler.

nexus-operations/update-workflow/handler/app.go
```diff
@@ -0,0 +1,65 @@
...
+	temporalnexus.TemporalOperationOptions[api.Input, api.Output]{
+		Name: api.IncrOperationName,
+		Start: func(
+			ctx context.Context,
+			nc temporalnexus.NexusClient,
+			input api.Input,
+			options temporalnexus.StartTemporalOperationOptions,
+		) (temporalnexus.TemporalOperationResult[api.Output], error) {
+			return temporalnexus.StartUpdateWorkflow[api.Output](ctx, nc, client.UpdateWorkflowOptions{
+				WorkflowID:   input.WorkflowID,
```
Quinn-With-Two-Ns, line 25: The SDK should require Update ID, like we require Workflow ID

nexus-operations/update-workflow/handler/app.go
```diff
@@ -0,0 +1,65 @@
...
+// CounterWorkflow is a long-running workflow that handles incr requests and stays open until it
+// receives the Done signal.
+func CounterWorkflow(ctx workflow.Context) (int, error) {
+	logger := workflow.GetLogger(ctx)
+	count := 0
+
+	if err := workflow.SetUpdateHandler(
+		ctx,
+		api.IncrUpdateName,
+		func(ctx workflow.Context) (api.Output, error) {
```
Quinn-With-Two-Ns: NIT: I would show an input to the update handler so users see that a parameter can be passed through and how to do it. It is pretty trivial, but still good to show so users can more easily adapt this sample for their use case that will likely need a parameter. For this sample I would just pass through the amount you want to increase the counter.

nexus-operations/update-workflow/caller/workflow.go
mavemuri: Good point - ill add both!

nexus-operations/update-workflow/handler/app.go
mavemuri: I was thinking about this and added the guard here 
https://github.com/temporalio/sdk-go/pull/2417/changes#diff-a0e4841a913b018fa07362d309ee10c01435a802c443044b14d9aaa360c2146eR186 

In regular UpdateWorkflow, an UpdateID is randomly generated if not supplied 
This keeps the experience consistent - for an operation that does workflow update, use a random id - which happens to be the request ID- as update ID 
If it retries due to a glitch, it will still be deterministic as the reqID is set on the nexusClient opts itself 
And it keeps experience similar for the user as well whether theyre using workflow update within same ns or across 

One thing im not 100% on is whether reqID itself can be empty - if yes, would need to add a check SDK side to generate random ID but remaining flow would be the same(set the updateID to requestID if empty)

nexus-operations/update-workflow/README.md
```diff
@@ -0,0 +1,121 @@
...
+
+There are two samples in this repo 
+1. Nexus Op UpdateWorkfow inside in a workflow. This is for cases where the Nexus Op will be called inside a workflow - see `caller` folder for details
+2. Standalone Nexus Op Update Workflow. This is for use-cases where the Nexus Op needs to be called from outside of a workflow. See the `standalone` folder for details 
+
+For both cases, the handler remains the same 
+
+### Running Standalone Nexus Operation UpdateWorkflow
+
+Ensure the temporal server has the required configurations. If required, start the dev server locally(v1.8.3 or later) with the dynamic config flags required for standalone Nexus operations:
```
mavemuri, line 17: v1.8.3 is tentative, not yet released - scheduled for end of this week

PR #504, opened 2026-07-07
Add Google ADK agent samples

DABH:
## What

Adds **Google ADK agent** samples under `googleadk/`, demonstrating the
[`go.temporal.io/sdk/contrib/googleadk`](https://pkg.go.dev/go.temporal.io/sdk/contrib/googleadk)
integration (shipped in temporalio/sdk-go#2439, released as `contrib/googleadk/v0.1.0`):
native ADK (`adk-go`) agents run durably on Temporal — the agent loop runs in a Workflow,
each model call runs as the `InvokeModel` Activity, and tools that do I/O run as durable
Activities.

Four scenarios, each with a `worker/` + `starter/` main, a README, and a `FakeModel`-driven
test that passes with no API key or network:

- **`googleadk/`** — basic agent: `AgentWorkflow` builds a native ADK `llmagent` with
  `googleadk.NewModel(...)`, exposes a `get_weather` Temporal activity to the agent via
  `googleadk.ActivityAsTool`, and drives the loop in-workflow with `NewContext(ctx)`.
- **`googleadk/multiagent/`** — a coordinator agent delegates to weather/jokes specialist
  `SubAgents` via ADK's `transfer_to_agent`; the whole tree runs in-workflow.
- **`googleadk/humanintheloop/`** — a sensitive tool calls `ctx.RequestConfirmation`; the
  workflow durably waits on a signal for the human decision, then resumes the agent with
  `googleadk.ConfirmationResponse` — one decision per resume pass, per the contrib docs'
  guidance.
- **`googleadk/chat/`** — a long-lived chat: each user message is a Workflow **Update**
  (`send-message`) that returns the answer directly (no polling), running on one shared ADK
  session; the workflow continues-as-new to bound history, carrying the session across the
  boundary via `googleadk.ExportSession` / `ImportSession`.

Also adds a `/googleadk/ @temporalio/sdk @temporalio/ai-sdk` CODEOWNERS entry, and
`@@@SNIPSTART` markers so the [Go integration docs page](https://github.com/temporalio/documentation/pull/4894)
can source its snippets from these samples.

Structure mirrors the existing contrib-plugin samples (e.g. `workflowstreams`): a root
workflow file + `worker/` and `starter/` mains, plus per-scenario subdirectories.

## Notes

- `go.mod` pins the released `go.temporal.io/sdk/contrib/googleadk v0.1.0` — no `replace`
  directives.
- Adds `workflowcheck.config.yaml` (wired into the Makefile's `workflowcheck` target)
  marking the contrib bridge's entry points (`runner.Run`, `ExportSession`,
  `PendingConfirmations`, `functiontool.New`) deterministic: the static checker can't see
  that `googleadk.NewContext` installs Temporal-deterministic time/UUID/task-fan-out
  providers and dispatches model/tool I/O as Activities. Determinism is replay-tested in
  the contrib package; the config exempts only those functions rather than whole loop
  bodies.
- To run the samples against a real model, set `GEMINI_API_KEY` (or `GOOGLE_API_KEY`) in
  the worker's environment; the tests themselves run offline via `googleadk.FakeModel`.


brianstrauch (commented):
Can you add @temporalio/ai-sdk to CODEOWNERS for this one?

googleadk/chat/starter/main.go
```diff
@@ -0,0 +1,59 @@
...
+			log.Fatalln("Unable to send message signal", err)
+		}
+		log.Printf("Sent message: %q", m)
+		// Give the agent time to answer before querying.
+		time.Sleep(2 * time.Second)
+
+		resp, err := c.QueryWorkflow(context.Background(), workflowID, "", chat.LatestAnswerQueryType)
+		if err != nil {
+			log.Fatalln("Unable to query latest answer", err)
+		}
```
brianstrauch: Updates are the better pattern for agent turns than signal + query since we can avoid polling

DABH: Done — refactored the chat scenario to use a Workflow Update (`send-message`) instead of signal + query. The update runs one agent turn on the shared session and returns the answer on the same call, so the starter no longer sleeps + polls a query. Turns are serialized (so concurrent updates can't interleave) and continue-as-new drains in-flight updates via `AllHandlersFinished` before exporting the session. (commit 85e66de)

DABH:
Thanks @brianstrauch — addressed the feedback and rebased:

- **Conflicts:** merged latest `main` (only `go.mod` conflicted; resolved + `go mod tidy`). PR is mergeable again.
- **CODEOWNERS:** added `/googleadk/ @temporalio/sdk @temporalio/ai-sdk`.
- **Agent turns → Updates:** the chat scenario now uses a `send-message` Workflow Update that returns the answer directly (no signal + query polling), with serialized turns and an `AllHandlersFinished` drain before continue-as-new. (85e66de)
- **Title/description:** updated to cover all four scenarios (basic, multi-agent, human-in-the-loop, chat), not just the first.

Still a draft pending the `contrib/googleadk` tagged release — the `go.mod` `replace` + version pin get dropped then.

brianstrauch:
@copilot resolve the merge conflicts in this pull request

.github/CODEOWNERS
```diff
@@ -5,3 +5,4 @@
 * @temporalio/sdk
 
 /workflowstreams/ @temporalio/sdk @temporalio/ai-sdk
+/googleadk/ @temporalio/sdk @temporalio/ai-sdk
```
brianstrauch:
```suggestion
/googleadk/       @temporalio/sdk @temporalio/ai-sdk
/workflowstreams/ @temporalio/sdk @temporalio/ai-sdk
```

googleadk/chat/starter/main.go
```diff
@@ -0,0 +1,59 @@
+package main
+
+import (
+	"context"
+	"log"
+
+	"go.temporal.io/sdk/client"
+	"go.temporal.io/sdk/contrib/envconfig"
+
+	chat "github.com/temporalio/samples-go/googleadk/chat"
```
brianstrauch: I don't think this alias is needed

brianstrauch (approved):

googleadk/chat/starter/main.go
DABH: Right — the package is already named `chat`; alias dropped. (7c243c7)

DABH:
Updates now that temporalio/sdk-go#2439 has merged:

- **Real dependency**: dropped the local sibling-checkout `replace`; `go.mod` now requires `go.temporal.io/sdk/contrib/googleadk v0.0.0-20260717224236-b77dd17ed4ce` (sdk-go `main`, includes the crypto bump). One caveat: the `go.temporal.io` vanity server doesn't serve the new module path yet (fix: temporalio/vanity-imports#23), so until that deploys there's a versioned `replace` to the GitHub path pinned at the same commit — buildable by CI and anyone, no local checkout needed. Once the vanity page is up (or a `contrib/googleadk` tag exists), deleting the replace is the only cleanup left.
- **HITL guidance**: the sample already resumed one decision per pass, but its README/comments suggested batching via `ConfirmationResponse(decisions...)` — the merged contrib docs now recommend against that for Activity-backed tools (batch resume order isn't replay-stable; upstream fix: google/adk-go#1169), so both are reworded to the one-decision-per-pass pattern.
- **CLA**: the two Copilot-authored conflict-resolution commits were recreated with identical trees under my authorship, so the CLA check should pass now.
- All `googleadk` sample tests pass against the real merged module; full repo builds clean.

DABH:
CI note: `build-and-test` newly reached workflowcheck on the googleadk packages — previously the local-checkout replace made them unbuildable in CI, so the analyzer never saw them. It flags ADK runner internals (channels/map-iteration/reflection) that are deterministic by construction under the contrib bridge (Temporal-deterministic time/UUID/fan-out seams, model+I/O as Activities; replay-tested in contrib). Added `workflowcheck.config.yaml` marking those specific entry points (`runner.Run`, `ExportSession`, `PendingConfirmations`, `functiontool.New`) deterministic and wired `-config` into the Makefile target — chose the config over inline `//workflowcheck:ignore` blocks so only those functions are exempt, not whole loop bodies. `workflowcheck ./...` is clean repo-wide.

PR #530, opened 2026-08-25
AI-471 Add Google ADK metrics sample

xumaple:
This PR adds a Google ADK sample that exports successful model-call and token-usage metrics through Temporal’s replay-aware Workflow metrics handler. It includes a Prometheus worker, starter, and API-key-free tests for exact metric values and missing usage metadata.


googleadk/metrics/workflow.go
```diff
@@ -0,0 +1,84 @@
+package metrics
```
DABH, line 1: No package doc, and no doc comments on `TaskQueue`, `ModelName` or `AgentWorkflow` — `grep -c '//'` returns 0 for all four new files, against 40/49/34 in `chat`, `humanintheloop` and `multiagent`'s `workflow.go`. Worth spending the lines here specifically, because the sample's one non-obvious decision is invisible as written: why `workflow.GetMetricsHandler(ctx)` and not a captured tally scope.

```suggestion
// Package metrics demonstrates recording Google ADK model token usage as
// Temporal Workflow metrics. An llmagent AfterModelCallback runs on the
// Workflow goroutine right after the googleadk.InvokeModel Activity resolves,
// so it can use workflow.GetMetricsHandler(ctx), whose emissions are suppressed
// during replay. Capturing a tally scope with sdktally.ScopeFromHandler would
// compile and then double-count on every replay.
package metrics
```

DABH (commented):
The design is right, and I checked the load-bearing part rather than assuming it: ADK runs `AfterModelCallback` synchronously on the Workflow goroutine (`base_flow.go:783` -> `:804`, no goroutine), so capturing `workflow.GetMetricsHandler(ctx)` at `workflow.go:24` is legal, and `.WithTags` preserves the replay-aware wrapper. Choosing that over `sdktally.ScopeFromHandler` is the one decision that makes this replay-safe, and `contrib/tally`'s `handler.go:54-56` warns about exactly the trap you avoided. Counter semantics are right too — `TemporalModel.GenerateContent` yields one aggregated response per turn, and Activity retries are invisible to the Workflow, so `model_calls` is genuinely one-per-turn with no retry inflation.

Two things that aren't tied to a line in this diff:

**The sample isn't registered anywhere.** The root `README.md:176-181` Google ADK bullet names all three existing scenarios and links them; `googleadk/README.md:23-37` lists four. Neither mentions `metrics/`, and this diff touches only the five files under `googleadk/metrics/`, so there is no path to this sample from any index. Extending the root bullet:

```
signal), a [continue-as-new chat](./googleadk/chat), and [model-usage
metrics](./googleadk/metrics).
```

and a fifth bullet in `googleadk/README.md`:

```markdown
- **[metrics/](metrics)** — records per-model-call token usage from
  `LLMResponse.UsageMetadata` onto Temporal's replay-aware Workflow metrics
  handler, exported via Prometheus.
```

**CI hasn't actually validated any of this.** `build-and-test` failed at `staticcheck` on a pre-existing SA1019 in `temporal-fixtures/rainbow-statuses/starter/main.go:47`, and since `Makefile:57` is `ci-build: staticcheck errcheck workflowcheck bins test`, the abort means errcheck, workflowcheck, bins and test never ran on this branch. I ran all five locally and they're green, so nothing is hiding — but the green-looking gate here is worth knowing about, and it'll keep failing for unrelated reasons until that fixture is fixed separately.

googleadk/metrics/workflow.go
```diff
@@ -0,0 +1,84 @@
...
+	"go.temporal.io/sdk/contrib/googleadk"
+)
+
+const (
+	TaskQueue = "google-adk-metrics"
+	ModelName = "gemini-2.5-flash"
+	agentName = "metrics-assistant"
+)
+
+func AgentWorkflow(ctx workflow.Context, prompt string) (string, error) {
```
DABH, line 24: No snippet markers, where all three sibling scenarios have them — `grep -rn "SNIPSTART" googleadk/` returns 14 hits across `workflow.go`, `multiagent/`, `chat/` and `humanintheloop/`, and zero under `metrics/`. `documentation/docs/develop/go/integrations/google-adk.mdx` snipsyncs seven of them today, so without markers, documenting this scenario later needs a second round-trip PR back into samples-go.

The sibling scenarios mark only `workflow.go`, not the worker/starter, so one pair here plus the matching `// @@@SNIPEND` after `:65`. A second pair around `afterModelMetricsCallback` (`:67-84`) is the higher-value snippet — the callback is the actual teaching unit.

```suggestion
// @@@SNIPSTART googleadk-metrics-workflow
func AgentWorkflow(ctx workflow.Context, prompt string) (string, error) {
```

googleadk/metrics/workflow.go
```diff
@@ -0,0 +1,84 @@
...
+	handler := workflow.GetMetricsHandler(ctx).WithTags(map[string]string{
+		"agent": agentName,
+		"model": ModelName,
+	})
+
+	root, err := llmagent.New(llmagent.Config{
```
DABH, line 30: There's no `google_adk_`-namespaced error signal, and the `responseErr != nil` branch at `:69` that looks like it provides one can't fire: `runAfterModelCallbacks` has a single call site (`base_flow.go:804`) and `:784-799` guarantees the error reaching it is nil — the error path diverts to `runOnModelErrorCallbacks`. I registered an always-erroring `model.LLM` to confirm: 10 `InvokeModel` attempts, Workflow fails with `googleadk.ModelError`, zero counters, neither branch hit.

Failures aren't invisible in production — the worker shares its tally scope with the SDK client, so `:9090` also serves `temporal_activity_execution_failed_total{activity_type="googleadk.InvokeModel"}` — but a metrics sample should show the framework-level signal. `OnModelErrorCallbacks` (`llmagent.go:239`, callback signature at `:378`) is the seam. Worth a third test driving an erroring model as well; `humanintheloop/workflow_test.go:72` is the sibling precedent for a negative case.

googleadk/metrics/workflow.go
```diff
@@ -0,0 +1,84 @@
...
+}
+
+func afterModelMetricsCallback(handler client.MetricsHandler) llmagent.AfterModelCallback {
+	return func(_ agent.Context, response *model.LLMResponse, responseErr error) (*model.LLMResponse, error) {
+		if responseErr != nil || response == nil {
+			return response, responseErr
+		}
+		handler.Counter("google_adk_model_calls").Inc(1)
+		if response.UsageMetadata == nil {
+			return response, nil
```
DABH: Returning the response terminates the callback chain. `base_flow.go:900-907` is `if callbackResponse != nil || callbackErr != nil { return callbackResponse, callbackErr }`, and `llmagent.go:231-233` documents it — callbacks run "until there's a callback that returns a non-nil LLMResponse or error. Then actual LLM response is replaced with the returned response/error." So an observe-only callback should return `nil, nil`; returning `response` means "replace the model's response with this one, and skip the rest."

Harmless in the sample as shipped, since it registers exactly one callback. The problem is pedagogical — this is written to be pasted into an existing ADK agent, and doing that silently kills that agent's other post-model hooks. I checked with a second callback registered after this one: it fired 0 times.

```suggestion
			return nil, nil
```

googleadk/metrics/workflow.go
```diff
@@ -0,0 +1,84 @@
...
+		}
+		usage := response.UsageMetadata
+		handler.Counter("google_adk_input_tokens").Inc(int64(usage.PromptTokenCount))
+		handler.Counter("google_adk_output_tokens").Inc(int64(usage.CandidatesTokenCount))
+		handler.Counter("google_adk_reasoning_tokens").Inc(int64(usage.ThoughtsTokenCount))
+		handler.Counter("google_adk_cached_input_tokens").Inc(int64(usage.CachedContentTokenCount))
+		handler.Counter("google_adk_total_tokens").Inc(int64(usage.TotalTokenCount))
+		return response, nil
```
DABH: Companion to the `:74` comment — same chain-termination, same fix.

```suggestion
		return nil, nil
```

googleadk/metrics/workflow_test.go
```diff
@@ -0,0 +1,80 @@
...
+	require.Equal(t, want, result)
+	return scope
+}
+
+func assertCounters(t *testing.T, snapshot ubertally.Snapshot, want map[string]int64) {
+	t.Helper()
+	got := make(map[string]int64, len(snapshot.Counters()))
+	for key, counter := range snapshot.Counters() {
+		got[key] = counter.Value()
+	}
```
DABH, line 100: `require.Equal` over the whole snapshot asserts these are the *only* counters in the scope. True today, but the suite hands the same handler to the activity and local-activity task handlers (`internal_workflow_testsuite.go:898`, `:1840`, `:2378`) — I reproduced `temporal_local_activity_total` and `temporal_activity_task_error` landing in the same snapshot under other configurations. Filtering by prefix is exactly as strong an assertion and doesn't break the day the SDK adds a metric (needs `"strings"` added to the imports):

```suggestion
	for key, counter := range snapshot.Counters() {
		if !strings.HasPrefix(key, "google_adk_") {
			continue
		}
		got[key] = counter.Value()
	}
```

googleadk/metrics/README.md
```diff
@@ -0,0 +1,28 @@
+# Google ADK model usage metrics
+
+This sample records successful Google ADK model calls and their input, output, reasoning, cached-input, and total token counts from `LLMResponse.UsageMetadata`. The ADK callback uses Temporal's Workflow metrics handler, which suppresses emissions during replay, and tags each counter with fixed agent and model names.
```
DABH: The six metric names appear nowhere in any markdown file in the repo — only in `workflow.go:72-81` and the test literals. For an observability sample the metric contract is the deliverable, so it's worth naming them here.

Worth flagging the naming divergence at the same time, because a reader has no way to discover it: `worker/main.go:41` wraps the scope in `sdktally.NewPrometheusNamingScope`, whose `Counter` appends `_total` when absent (`contrib/tally@v0.2.0/prometheus.go:47-51`), and `PrometheusSanitizeOptions` rewrites dots and dashes to underscores in tag values. I reproduced the worker's exact wiring in-process and scraped it:

```
google_adk_model_calls_total{agent="metrics_assistant",model="gemini_2_5_flash",workflow_type="AgentWorkflow"} 1
```

So none of the names in the test are the deployed names, and the SDK adds `workflow_type` on top. Suggested replacement for this paragraph:

```suggestion
This sample records successful Google ADK model calls and their token counts from `LLMResponse.UsageMetadata`. The ADK callback emits through Temporal's Workflow metrics handler, which suppresses emissions during replay, and tags each counter with fixed agent and model names.

| Counter (as scraped) | Source field |
| --- | --- |
| `google_adk_model_calls_total` | one per successful model turn |
| `google_adk_input_tokens_total` | `PromptTokenCount` |
| `google_adk_output_tokens_total` | `CandidatesTokenCount` |
| `google_adk_reasoning_tokens_total` | `ThoughtsTokenCount` |
| `google_adk_cached_input_tokens_total` | `CachedContentTokenCount` (a subset of input) |
| `google_adk_total_tokens_total` | `TotalTokenCount` (also counts tool-use prompt tokens, which this sample never records) |

`NewPrometheusNamingScope` adds the `_total` suffix and sanitizes tag values, and the SDK adds a `workflow_type` tag, so a scraped line reads `google_adk_model_calls_total{agent="metrics_assistant",model="gemini_2_5_flash",workflow_type="AgentWorkflow"} 1`.

Replay suppression makes these at-least-once rather than exactly-once — a Workflow task that fails before completing re-executes its non-replayed segment and can record again. Treat them as usage signals, not billing records.
```

googleadk/metrics/README.md
```diff
@@ -0,0 +1,28 @@
...
+In another terminal, start the Workflow:
+
+```sh
+go run googleadk/metrics/starter/main.go
+```
+
+Inspect the Prometheus endpoint:
+
+```sh
+curl http://127.0.0.1:9090/metrics | rg 'google_adk_'
```
DABH: `rg` is the only ripgrep invocation in any markdown file in the repo (`grep -rn "| rg " --include='*.md' .` returns this one hit) and it isn't a documented prerequisite anywhere. The pattern also won't match what's actually served, per the `:3` comment — the counters carry a `_total` suffix.

```suggestion
curl -s http://127.0.0.1:9090/metrics | grep google_adk_
```

googleadk/metrics/workflow.go
xumaple: Disagree. Samples should be comment free as a baseline.

googleadk/metrics/workflow.go
xumaple: Done

googleadk/metrics/workflow.go
xumaple: Done. Added an OnModelErrorCallback counter and a failing-model test.

googleadk/metrics/workflow.go
xumaple: Done. The observation callback now returns nil, nil.

googleadk/metrics/workflow.go
xumaple: Done.

googleadk/metrics/workflow_test.go
xumaple: Done. The assertion now filters to google_adk_ counters.

googleadk/metrics/README.md
xumaple: Done. Added the scraped counter names, tags, and at-least-once semantics.

googleadk/metrics/README.md
xumaple: Done.

xumaple:
Added the sample to both Google ADK indexes. The scoped test, vet, build, staticcheck, errcheck, and workflowcheck pass; the full gate still stops at the pre-existing rainbow-statuses SA1019 failure.

DABH (approved):
LGTM

googleadk/metrics/workflow.go
DABH: This is probably worth chatting about more offline but I'm ok with this and won't let this block.  I guess the point of "why workflow.GetMetricsHandler(ctx) and not a captured tally scope" might still be worth documenting somewhere but possibly in the plugin or elsewhere in the samples vs. at the package level scope

tconley1428 (approved):

